PR #2, opened 2020-01-31
Add .travis.yml

Sean-Der:
Add everything needed to run tests and generate code coverage


.travis.yml
```diff
@@ -0,0 +1,120 @@
...
+addons:
+  apt:
+    packages:
+      - gdb
+
+script:
+  - export AWS_KVS_LOG_LEVEL=2
+  - make
+  - ulimit -c unlimited -S
+  - timeout --signal=SIGABRT 20m ./tst/producer_test
```
MushMal: Might be too tight. We will adjust this later.

Sean-Der: :+1: I just moved to two hours.

Issue #11, opened 2020-02-24
cross-compile branch fail.


awsa2ron:
Hi,
I have setup a new ubuntu18.04 docker and try to make the "cross-compile" branch. But failed.
I followed those steps:
```
# Cross-compilation to ARM, no tests are run
sudo apt-get install gcc-arm-linux-gnueabi
sudo apt-get install g++-arm-linux-gnueabi
sudo apt-get install binutils-arm-linux-gnueabi

#for Linux
export CC=arm-linux-gnueabi-gcc CXX=arm-linux-gnueabi-g++ AR=arm-linux-gnueabi-ar
mkdir build && cd build
cmake .. -DBUILD_TEST=TRUE -DBUILD_OPENSSL=TRUE \
         -DBUILD_OPENSSL_PLATFORM=linux-generic32 \
         -DBUILD_LIBCURL_HOST_PLATFORM=x86_64-unknown-linux-gnu \
         -DBUILD_LIBCURL_DESTINATION_PLATFORM=arm-unknown-linux-uclibcgnueabi

make
```
Here is the log:
```
Scanning dependencies of target kvsCommonCurl
[ 52%] Building C object CMakeFiles/kvsCommonCurl.dir/src/source/Common/Auth.c.o
In file included from /usr/include/openssl/sha.h:13:0,
                 from /home/aaron/producer-c/src/source/Common/Include_i.h:16,
                 from /home/aaron/producer-c/src/source/Common/Auth.c:5:
/usr/include/openssl/e_os2.h:13:11: fatal error: openssl/opensslconf.h: No such file or directory
 # include <openssl/opensslconf.h>
           ^~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
CMakeFiles/kvsCommonCurl.dir/build.make:62: recipe for target 'CMakeFiles/kvsCommonCurl.dir/src/source/Common/Auth.c.o' failed
make[2]: *** [CMakeFiles/kvsCommonCurl.dir/src/source/Common/Auth.c.o] Error 1
CMakeFiles/Makefile2:106: recipe for target 'CMakeFiles/kvsCommonCurl.dir/all' failed
make[1]: *** [CMakeFiles/kvsCommonCurl.dir/all] Error 2
Makefile:83: recipe for target 'all' failed
make: *** [all] Error 2
```

MushMal:
Looks like the issue with cross-compiling the libCurl. We are working on it currently.

[...]

PR #14, opened 2020-03-01
fix openssl --with-ssl compile option

awsa2ron:
Issue #11

Fix --with-ssl compile option for openssl, and make a lib folder with libssl.a and libcrypto.a links inside it, which are required by cURL.


MushMal:
@awsa2ron could you please give some details as to why you need statically linked openssl/libCrypto? For vast majority of the cases, we would discourage static linkage.

awsa2ron:
Thank you for remind. I just want to fix a error when I compile it. But looks I go to the wrong dirction.

PR #35, opened 2020-03-13
use caching endpoint by default in samples

chehefen:
*Issue #, if available:*

*Description of changes:*


chehefen:
> Why are we using caching endpoint in the samples?
should'nt it be used by default across all applications? Otherwise we add extra latency by calling described streams every time we create new putMedia.

PR #36, opened 2020-03-16
Add mbedtls

chehefen:
*Issue #, if available:*

*Description of changes:*


CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ project(KinesisVideoProducerC LANGUAGES C)
 option(ADD_MUCLIBC "Add -muclibc c flag" OFF)
 option(BUILD_DEPENDENCIES "Whether or not to build depending libraries from source" ON)
 option(USE_OPENSSL "Use openssl as crypto library" ON)
+option(USE_MBEDTLS "Use openssl as crypto library" OFF)
```
zhiyua-git: use mbedTLS?

CMakeLists.txt
```diff
@@ -63,10 +64,21 @@ if(BUILD_DEPENDENCIES)
 
   message(STATUS "Begin building dependencies.")
 
+  count_true(ENABLED_SSL_OPTION_COUNT
+    USE_OPENSSL
+    USE_MBEDTLS)
+
+  if(ENABLED_SSL_OPTION_COUNT GREATER "1")
```
zhiyua-git: If my input is just cmake -DUSE_MBEDTLS=TRUE .., would it fail? If it fails, is the error message clear enough to let me know that I need to disable openssl by -DUSE_OPENSSL=FALSE?

src/source/Common/Include_i.h
```diff
@@ -13,9 +13,15 @@ extern "C" {
 ////////////////////////////////////////////////////
 // Project include files
 ////////////////////////////////////////////////////
+#if defined(KVS_USE_OPENSSL)
 #include <openssl/sha.h>
 #include <openssl/hmac.h>
 #include <openssl/evp.h>
+#elif defined(KVS_USE_MBEDTLS)
+#include "mbedtls/sha256.h"
+#include <mbedtls/md.h>
```
zhiyua-git: minor: shouldn't both be <mbedtls/*.h>?

src/source/Common/AwsV4Signer.c
```diff
@@ -1070,3 +1024,100 @@ PCHAR getRequestVerbString(HTTP_REQUEST_VERB verb)
 
     return NULL;
 }
+
+#if defined(KVS_USE_OPENSSL)
```
MushMal: I strongly discourage usage of conditional compilations in this manner. 

Recommend using a separate header that hides the #ifdef under a macro and use those macros in the single code body. 

You only need to abstract two API calls. sha256 and hmac.

src/source/Common/AwsV4Signer.h
```diff
@@ -78,6 +78,10 @@ extern "C" {
 // URI-encoded backslash value
 #define URI_ENCODED_FORWARD_SLASH               "%2F"
 
+#define SHA256_DIGEST_LENGTH                	32
```
MushMal: Very minor: line up the values.

src/source/Common/Include_i.h
MushMal: Using wildcards is a bad idea

zhiyua-git: Didn't mean wildcard, I meant, shouldn't it be both 
#include "mbedtls/sha256.h"
#include "mbedtls/md.h"
or 
#include <mbedtls/sha256.h>
#include <mbedtls/md.h>

CMakeLists.txt
chehefen: all our variables are capitalized though.

src/source/Common/AwsV4Signer.h
chehefen: it does line up in clion.

CMakeLists.txt
MushMal: +1 - capitalized is used everywhere

src/source/Common/Include_i.h
MushMal: Oh, I see. Yeah, I think angle brackets are the right way to go here

Issue #41, opened 2020-03-24
0x52000017 error encountered while creating a stream with new name - KvsVideoOnlySample


disa6302:
While creating a stream with a new name, the VideoOnlySample encounters a HTTP 404 error followed by getStreamingEndpointCurlHandler resulting in 0x52000017 error

disa6302:
[error_with_new_stream_name.txt](https://github.com/awslabs/amazon-kinesis-video-streams-producer-c/files/4372647/error_with_new_stream_name.txt)

PR #42, opened 2020-03-24
Introducing different levels of caching.

MushMal:
Fixing the video only application
Reving up the SDK major version due to public API change!!!
Adding unit test targeting the behavior

Issue: https://github.com/awslabs/amazon-kinesis-video-streams-producer-c/issues/41

Issue #41 

*Description of changes:*


src/source/Common/RequestInfo.c
```diff
@@ -299,6 +299,10 @@ SERVICE_CALL_RESULT getServiceCallResultFromHttpStatus(UINT32 httpStatus)
         case SERVICE_CALL_NOT_AUTHORIZED:
         case SERVICE_CALL_NOT_IMPLEMENTED:
         case SERVICE_CALL_INTERNAL_ERROR:
+        case SERVICE_CALL_REQUEST_TIMEOUT:
```
MushMal: This is not really related to this PR but we need this for the upcoming changes for the intermittent producer and this is in general a good correctness enhancement

src/source/Common/Version.h
```diff
@@ -13,7 +13,7 @@ extern "C" {
 /**
  * IMPORTANT!!! This is the current version of the SDK which needs to be maintained
  */
-#define AWS_SDK_KVS_PRODUCER_VERSION_STRING             (PCHAR) "2.1.0"
+#define AWS_SDK_KVS_PRODUCER_VERSION_STRING             (PCHAR) "3.0.0"
```
MushMal: NOTE. This is a major version change due to public API changes

tst/ProducerApiTest.cpp
```diff
@@ -9,12 +9,12 @@ TEST_F(ProducerApiTest, fileLoggerApiTest)
 {
     PClientCallbacks pClientCallbacks = NULL;
     PCHAR testLogDir = ((PCHAR) "/tmp");
-    PCHAR longPath = (PCHAR) MEMALLOC(MAX_PATH_LEN + 2);
+    CHAR longPath[MAX_PATH_LEN + 2];
```
MushMal: No need to heap alloc

src/source/CurlApiCallbacks.c
```diff
@@ -42,6 +42,9 @@ STATUS createCurlApiCallbacks(PCallbacksProvider pCallbacksProvider, PCHAR regio
     // Endpoint period will be the same for all streams
     pCurlApiCallbacks->cacheUpdatePeriod = endpointCachingPeriod;
 
+    // API call caching type
+    pCurlApiCallbacks->cacheType = cacheType;
```
MushMal: Storing the caching type so we can use it later as we end up with a single callback entries few lines below

src/source/CurlApiCallbacks.c
```diff
@@ -52,17 +55,10 @@ STATUS createCurlApiCallbacks(PCallbacksProvider pCallbacksProvider, PCHAR regio
...
-        pCurlApiCallbacks->apiCallbacks.describeStreamFn = describeStreamCachingCurl;
-        pCurlApiCallbacks->apiCallbacks.getStreamingEndpointFn = getStreamingEndpointCachingCurl;
-        pCurlApiCallbacks->apiCallbacks.tagResourceFn = tagResourceCachingCurl;
-    } else {
-        pCurlApiCallbacks->apiCallbacks.createStreamFn = createStreamCurl;
-        pCurlApiCallbacks->apiCallbacks.describeStreamFn = describeStreamCurl;
-        pCurlApiCallbacks->apiCallbacks.getStreamingEndpointFn = getStreamingEndpointCurl;
-        pCurlApiCallbacks->apiCallbacks.tagResourceFn = tagResourceCurl;
-    }
+    pCurlApiCallbacks->apiCallbacks.createStreamFn = createStreamCachingCurl;
```
MushMal: Will always call the caching version, which, in turn will run the logic

src/source/CurlApiCallbacks.c
```diff
@@ -1034,31 +1030,86 @@ STATUS createStreamCachingCurl(UINT64 customData, PCHAR deviceName, PCHAR stream
...
     CHK(pCurlApiCallbacks != NULL && pCurlApiCallbacks->pCallbacksProvider != NULL && pServiceCallContext != NULL, STATUS_INVALID_ARG);
     pCallbacksProvider = pCurlApiCallbacks->pCallbacksProvider;
 
-    DLOGV("No-op CreateStream API call");
-
     streamHandle = (STREAM_HANDLE) pServiceCallContext->customData;
 
+    // We check whether we have already made the call by checking
+    // for the presence of the endpoint in the cache.
+    switch(pCurlApiCallbacks->cacheType) {
```
MushMal: Similar pattern for all of the caching APIs. We check the caching logic first. If none then we just run the main function. If endpoint only (with the exception of the get endpoint API call) we emulate always. If it's cache all API calls, we look at the pre-cached endpoint and it's TTL. If all valid then we will emulate the calls, meaning that we had run these before.

src/source/CurlApiCallbacks.c
```diff
@@ -1896,25 +2025,83 @@ STATUS tagResourceCachingCurl(UINT64 customData, PCHAR streamArn, UINT32 tagCoun
...
+        case API_CALL_CACHE_TYPE_NONE:
+            emulateApiCall = FALSE;
+            break;
+
+        case API_CALL_CACHE_TYPE_ENDPOINT_ONLY:
+            emulateApiCall = TRUE;
+            break;
+
+        case API_CALL_CACHE_TYPE_ALL:
+            pCallbacksProvider->clientCallbacks.lockMutexFn(pCallbacksProvider->clientCallbacks.customData,
```
chehefen: Do we need all these? tagResourceResultEvent just take a stream handle.

src/source/CurlApiCallbacks.c
```diff
@@ -1896,25 +2025,83 @@ STATUS tagResourceCachingCurl(UINT64 customData, PCHAR streamArn, UINT32 tagCoun
...
+            if (retStatus == STATUS_HASH_KEY_NOT_PRESENT) {
+                emulateApiCall = FALSE;
+
+                // Reset the status if not found
+                retStatus = STATUS_SUCCESS;
+            } else {
+                pEndpointTracker = (PEndpointTracker) value;
+                curTime = pCallbacksProvider->clientCallbacks.getCurrentTimeFn(pCallbacksProvider->clientCallbacks.customData);
+
+                if (pEndpointTracker == NULL ||
```
chehefen: if cache has expired, tag resource wont get new endpoint.

src/source/CurlApiCallbacks.c
```diff
@@ -1271,12 +1322,68 @@ STATUS describeStreamCachingCurl(UINT64 customData, PCHAR streamName, PServiceCa
...
+        case API_CALL_CACHE_TYPE_NONE:
+            emulateApiCall = FALSE;
+            break;
+
+        case API_CALL_CACHE_TYPE_ENDPOINT_ONLY:
+            emulateApiCall = TRUE;
+            break;
+
+        case API_CALL_CACHE_TYPE_ALL:
+            pCallbacksProvider->clientCallbacks.lockMutexFn(pCallbacksProvider->clientCallbacks.customData,
```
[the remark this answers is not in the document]
MushMal: It can be done, a little ugly but should be possible. Let me do that.

src/source/CurlApiCallbacks.c
MushMal: Not sure what you mean. This block checks what caching level needs to be done and whether we need to call the main function to actually perform the call

src/source/CurlApiCallbacks.c
MushMal: We use the presence of the endpoint in the cache as an indicator whether the other calls have been made or not as the state machine is pretty much sequential. Get endpoint is the last in the sequence of FE API calls. Next API call is the actual PutMedia call

src/include/com/amazonaws/kinesis/video/cproducer/Include.h
```diff
@@ -241,6 +241,20 @@ struct __ApiCallbacks {
...
 
+/**
+ * Type of caching implementation
+ */
+typedef enum {
+    // No caching is specified, will make API calls
+    API_CALL_CACHE_TYPE_NONE,
+
+    // The stream should be pre-existing as DescribeStream API call will be emulated
+    API_CALL_CACHE_TYPE_ENDPOINT_ONLY,
```
disa6302: Considering we did not have endpoint and cache type all before, why do we have it now? Just trying to understand the context here.

MushMal: So, previously we had two behaviors controlled by a BOOL - either no caching at all or Endpoint ONLY caching. The third option that this adds is the caching ALL of the API calls which is necessary if the stream is not pre-created.

src/source/CurlApiCallbacks.c
```diff
@@ -2227,3 +2272,76 @@ STATUS findRequestWithUploadHandle(UPLOAD_HANDLE uploadHandle, PCurlApiCallbacks
...
+    }
+
+    if (endpointsLocked) {
+        pCallbacksProvider->clientCallbacks.unlockMutexFn(pCallbacksProvider->clientCallbacks.customData,
+                                                          pCurlApiCallbacks->cachedEndpointsLock);
+    }
+
+    LEAVES();
+    return retStatus;
+}
```
chehefen: no \n at end of file

src/include/com/amazonaws/kinesis/video/cproducer/Include.h
```diff
@@ -241,6 +241,20 @@ struct __ApiCallbacks {
 };
 typedef struct __ApiCallbacks* PApiCallbacks;
 
+/**
+ * Type of caching implementation
```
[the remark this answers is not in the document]
MushMal: Good point. Let me add more context in the comments. Will send a quick update shortly

src/include/com/amazonaws/kinesis/video/cproducer/Include.h
```diff
@@ -241,6 +241,20 @@ struct __ApiCallbacks {
 };
 typedef struct __ApiCallbacks* PApiCallbacks;
 
+/**
+ * Type of caching implementation
+ */
+typedef enum {
+    // No caching is specified, will make API calls
+    API_CALL_CACHE_TYPE_NONE,
```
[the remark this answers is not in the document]
MushMal: We should always prefix the enum members with the full enum name. This is needed due to flat namespace in C and we also chose this convention.

src/include/com/amazonaws/kinesis/video/cproducer/Include.h
```diff
@@ -241,6 +241,20 @@ struct __ApiCallbacks {
...
+ */
+typedef enum {
+    // No caching is specified, will make API calls
+    API_CALL_CACHE_TYPE_NONE,
+
+    // The stream should be pre-existing as DescribeStream API call will be emulated
+    API_CALL_CACHE_TYPE_ENDPOINT_ONLY,
+
+    // Call the APIs and cache the result
+    API_CALL_CACHE_TYPE_ALL,
```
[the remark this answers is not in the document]
MushMal: Let me add it in the comments. ALL is the same as NONE for the first call. It caches the info after the call. While, the endpoint only emulates the call always for non-endpoint API calls.

tst/CallbacksProviderApiTest.cpp
```diff
@@ -18,7 +18,7 @@ TEST_F(CallbacksProviderApiTest, createDefaultCallbacksProvider_variations)
                                                              mCaCertPath,
                                                              NULL,
                                                              TEST_USER_AGENT,
-                                                             FALSE,
+                                                             API_CALL_CACHE_TYPE_NONE,
```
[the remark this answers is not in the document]
MushMal: These are just the negative cases. I will add a case for ALL type too. The functionality test is in ProducerApiCallCacheTest.cpp

src/include/com/amazonaws/kinesis/video/cproducer/Include.h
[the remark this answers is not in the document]
sharadrj: Got it. 
IMO, it would be much simpler for SDK users if we they don't have to understand the underlying behavior and can just focus on the things they care about. So, instead of having modes to indicate API caching which is internal detail, the API can just differentiate between whether the customer has pre-created streams or not.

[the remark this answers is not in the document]
MushMal: True, this was what I thought too. The problem is that we have multiple public API entry points. We have a choice to either add another boolean to the public APIs (which is very poor API design at that stage) or permute the APIs by creating another set of APIs which would encompass the behavior. I think this would be the least of all evil.

tst/ProducerClientBasicTest.cpp
```diff
@@ -914,7 +924,7 @@ TEST_F(ProducerClientBasicTest, createStreamStreamUntilTokenRotationStopSyncFree
     EXPECT_EQ(STATUS_SUCCESS, createRealtimeVideoStreamInfoProvider(streamName, TEST_RETENTION_PERIOD, TEST_STREAM_BUFFER_DURATION, &pStreamInfo));
     pStreamInfo->streamCaps.nalAdaptationFlags = NAL_ADAPTATION_FLAG_NONE;
     EXPECT_EQ(STATUS_SUCCESS, createAbstractDefaultCallbacksProvider(TEST_DEFAULT_CHAIN_COUNT,
```
disa6302: Shouldn't we be adding tests for CACHE_TYPE_ALL as well?

MushMal: We do have coverage of the permutations in other tests. These functionality tests target other areas of functionality and permuting them at the higher level will produce a combinatorial explosion. We already have multiple dimensions that we target. Perhaps adding more might not make sense for now.

disa6302:
Looks good to me! Approved!

PR #48, opened 2020-04-02
cpd size is integer not pointer

awsa2ron:
cpd size is an integer not pointer


MushMal:
Re-published https://github.com/awslabs/amazon-kinesis-video-streams-producer-c/pull/49

PR #62, opened 2020-05-07
Windows Support

Sean-Der:
This is broken, but WIP

### DIFF
--- a/.github/build_windows.bat
+++ b/.github/build_windows.bat
@@ -1,0 +1,6 @@
+call "C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC\Auxiliary\Build\vcvars64.bat"
+mkdir build
+cd build
+cmd.exe /c cmake -G "NMake Makefiles" ..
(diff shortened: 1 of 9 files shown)


disa6302:
The bat script in windows would be used by customers as well, right? 

If that is the case, should we move the file outside .github and add instructions in the README.md?

Sean-Der:
I wasn't planning for customers to use @disa6302. Personally I would rather users invoke the commands themselves!

I want to have this all inline in the `.travis.yml` but I couldn't get bash on Windows to execute :/

disa6302:
@Sean-Der,  I am good with that! As long as we list out the dependencies with the version information we tested with (maybe) for running our SDKs on Windows in the cpp SDK, it should be good.

Sean-Der:
@disa6302 agree, after this I want to go back and every README will have a `building on Windows` note.

My end goal is that everything should have the same exact flow, but then we have little notes (If you are building on Windows/Mac/Linux do this!) 

I think we will see more stability/less config drift that way! Maybe I am wrong, but that is my hope at least :)

PR #67, opened 2020-05-13
Enhance build system for users

lherman-cs:
Changes:
* No longer treat PIC specially
* Add install steps
* Add pc files


lherman-cs:
Updated with PIC dependency pointing to a specific commit hash

lherman-cs:
@MushMal PTAL

PR #77, opened 2020-05-28
Outsourcing the file logging functionality into common. Adding new co…

MushMal:
…mmon API and tests

Issue #https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/289

*Description of changes:*


chehefen:
I think the file logging functionality should be put into PIC. webrtc client does not depend on producer common and and now it will have to do that to get the file logging function.

disa6302:
@chehefen , 

We do include <com/amazonaws/kinesis/video/common/Include.h> in the webrtc Include.h.

MushMal:
> I think the file logging functionality should be put into PIC. webrtc client does not depend on producer common and and now it will have to do that to get the file logging function.

WebRTC does include Common library so it will pick it up automatically. I think it should be OK

PR #78, opened 2020-05-29
[Issue-65] Add command line parser

### LINKED ISSUE #65: Have a common function to parse command line options
Currently, all the samples mandate sequence of command line arguments which makes the code less maintainable over time. Using getopt() is not the most direct option because Windows does not support it. We need to have a utility of sorts that can be used by cpp sdk and producer sdk to parse command line options

disa6302:
*Issue #, if available: #65 *

*Description of changes:*
- Currently, we require the command line options to be in a specific order. This creates a hassle because we will require all options to set a specific option. To avoid any platform dependency, this PR aims at creating a custom command line option parser. 
- Modified samples to move to this approach

Note: The parser mandates requirement of prefix `--` to the option.

Resolves #65 

### DIFF
--- a/samples/KvsAacAudioVideoStreamingSample.c
+++ b/samples/KvsAacAudioVideoStreamingSample.c
@@ -141,6 +141,16 @@ CleanUp:
     return (PVOID) (ULONG_PTR) retStatus;
 }
 
+void helper() {
+    printf("List of available options:\n" \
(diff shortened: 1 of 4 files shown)


disa6302:
Closing this PR for now. Will revisit when I can get to this.

PR #88, opened 2020-06-18
Add mbedTLS support for common LWS

lherman-cs:
This is required to enable mbedTLS support for WebRTC SDK, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/121.

### DIFF
--- a/CMake/Dependencies/libwebsockets-CMakeLists.txt
+++ b/CMake/Dependencies/libwebsockets-CMakeLists.txt
@@ -12,6 +12,12 @@ else()
   set(LWS_WITH_SHARED 1)
 endif()
 
+if (USE_MBEDTLS)
+  set(LWS_WITH_MBEDTLS ON)
+else()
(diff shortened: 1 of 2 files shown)


CMake/Dependencies/libwebsockets-CMakeLists.txt
```diff
@@ -12,6 +12,11 @@ else()
   set(LWS_WITH_SHARED 1)
 endif()
 
+set(LWS_WITH_MBEDTLS OFF)
+if (USE_MBEDTLS)
+  set(LWS_WITH_MBEDTLS ON)
+endif()
```
MushMal, line 18: Why not if/else?

lherman-cs: No particular reason really. But, yea, I think if else is more readable too. I'll update it.

PR #130, opened 2020-10-29
Update issue templates

hassanctech:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,47 @@
...
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Desktop (please complete the following information):**
+ - OS: [e.g. iOS]
+ - Browser [e.g. chrome, safari]
+ - Version [e.g. 22]
+
+**Smartphone (please complete the following information):**
```
MushMal, line 40: I would rather remove this section as most of the C producer users are not mobile clients. Instead, I would add something about the media pipeline

**Media pipeline:**
 - What is the media source (GStreamer, custom, etc). What is the upstream encoder/pipeline?
 - What is the media type, configuration
 - For Video - what's the codec information (Annex-B, AvCC, raw frames), CPD present, etc..
 - For Audio - what's the audio configuration (CPD generated, etc)..
 - Intermittent/continuous streaming?

MushMal (approved):
Thanks for getting this done!

.github/ISSUE_TEMPLATE/bug_report.md
hassanctech: Yeah good point, Agreed, I'm going to update this.

PR #151, opened 2020-12-14
Adding OpenSSL thread and mutex callbacks

MushMal:
Issue #https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/1000#issuecomment-744675653

Description of changes:

Note yet entirely sure whether this will have any impact but seems in some cases the OpenSSL would require the callback for thread id and create/lock/unlock/release locks functionality. This is experimental.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,7 @@ option(MEMORY_SANITIZER "Build with MemorySanitizer." OFF)
 option(THREAD_SANITIZER "Build with ThreadSanitizer." OFF)
 option(UNDEFINED_BEHAVIOR_SANITIZER "Build with UndefinedBehaviorSanitizer." OFF)
 option(ALIGNED_MEMORY_MODEL "Aligned memory model ONLY." OFF)
+option(SET_SSL_CALLBACKS "Set SSL thread and lock callbacks." OFF)
 
 set(CMAKE_MACOSX_RPATH TRUE)
 
@@ -209,8 +210,14 @@ if(ALIGNED_MEMORY_MODEL)
     add_definitions(-DALIGNED_MEMORY_MODEL)
 endif()
 
+if(SET_SSL_CALLBACKS)
+    add_definitions(-DSET_SSL_CALLBACKS)
+endif()
+
+
 # use openssl by default
 set(CPRODUCER_COMMON_TLS_OPTION KVS_USE_OPENSSL)
+
 set(PRODUCER_CRYPTO_LIBRARY
         OpenSSL::Crypto
         OpenSSL::SSL)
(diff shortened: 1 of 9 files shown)


MushMal:
Attached article which is next to impossible to decipherr: https://www.openssl.org/docs/man1.0.2/man3/threads.html

disa6302 (approved):
The changes itself looks good. I entirely do not understand what this does though. Can you add some explanation in the PR itself detailing what this exactly solves? I checked the ticket this references, but I am missing some context I guess.

MushMal:
> The changes itself looks good. I entirely do not understand what this does though. Can you add some explanation in the PR itself detailing what this exactly solves? I checked the ticket this references, but I am missing some context I guess.

Sure. From reading up whatever I could muster, it seems that the OpenSSL could have issues with some shared resources (not sure which ones) and it seems that some implementations/compile time might actually have an undefined behavior (crash) when there is a parallel contexts. From what I can understand, some implementations/versions/compiles would not have this issue while others might. There seems to be a compile time definition OPENSSL_THREADS that controls it.

Basically, the init functionality will call a few of the APIs

    ```
CRYPTO_set_locking_callback(callback);
    CRYPTO_THREADID_set_callback(callback);
    CRYPTO_set_dynlock_create_callback(callback);
    CRYPTO_set_dynlock_lock_callback(callback);
    CRYPTO_set_dynlock_destroy_callback(callback);

```

which would set the callbacks that the OpenSSL would call to perform locking in this example. We use our own locks which are platform independent that can perform the locking.

The important piece here is to let these changes to bake for a while. I would like to get the customer that reported a crash to give this a try before we push these through. The issue they had was related to WebRTC. Both libCurl and libWebSockets (by default) will use OpenSSL so this is the reason I put this functionality into the common so we can use in both of the projects.

hassanctech:
> > The changes itself looks good. I entirely do not understand what this does though. Can you add some explanation in the PR itself detailing what this exactly solves? I checked the ticket this references, but I am missing some context I guess.
> 
> Sure. From reading up whatever I could muster, it seems that the OpenSSL could have issues with some shared resources (not sure which ones) and it seems that some implementations/compile time might actually have an undefined behavior (crash) when there is a parallel contexts. From what I can understand, some implementations/versions/compiles would not have this issue while others might. There seems to be a compile time definition OPENSSL_THREADS that controls it.
> 
> Basically, the init functionality will call a few of the APIs
> 
> ```
> ```
> ```
> 
> CRYPTO_set_locking_callback(callback);
> CRYPTO_THREADID_set_callback(callback);
> CRYPTO_set_dynlock_create_callback(callback);
> CRYPTO_set_dynlock_lock_callback(callback);
> CRYPTO_set_dynlock_destroy_callback(callback);
> 
> ```
> 
> which would set the callbacks that the OpenSSL would call to perform locking in this example. We use our own locks which are platform independent that can perform the locking.
> 
> The important piece here is to let these changes to bake for a while. I would like to get the customer that reported a crash to give this a try before we push these through. The issue they had was related to WebRTC. Both libCurl and libWebSockets (by default) will use OpenSSL so this is the reason I put this functionality into the common so we can use in both of the projects.
> ```

Agreed we should wait for the customer that reported the crash to see if their issue was resolved before pushing this through.

MushMal:
Perhaps, one thing we should also do is to control this via our own compile time flag. What do people think?

PR #167, opened 2021-01-07
Fix tsan issues

hassanctech:
This doesn't address all the issues reported by TSAN but the ones addressed so far are:

(1) In getStreamingEndpointCachingCurl we were not checking if a mutex was locked before unlocking, check added.
(2) In T1 we call `createDeviceCurl` which grabs the `activeRequestsLock` then we call `createDeviceResultEvent` in PIC which calls `createDeviceResult` which acquired the client lock.  In T2 in `putFrame` we first acquire the stream lock, then the client lock.  Next we invoke the callback in producer-c `streamLatencyPressureFn` which eventually calls `getStreamingEndpointCurl` which now acquires the `activeRequestsLock`.  This could result in deadlock and was the *only* potential deadlock reported by TSAN.  I think we never hit this because no one uses create Device.  Nonetheless to remedy this I've removed the acquisition of `activeRequestsLock` by T1 in `createDeviceCurl`

### DIFF
--- a/CMake/Dependencies/libkvspic-CMakeLists.txt
+++ b/CMake/Dependencies/libkvspic-CMakeLists.txt
@@ -7,7 +7,7 @@ include(ExternalProject)
 # clone repo only
 ExternalProject_Add(libkvspic-download
 	GIT_REPOSITORY    https://github.com/awslabs/amazon-kinesis-video-streams-pic.git
-    	GIT_TAG           76223690da1329d733309e2a52575d6a98a68981
+    	GIT_TAG           988b1ab392b50f0ad573cc1a7183299f08425597
     	SOURCE_DIR        "${CMAKE_CURRENT_BINARY_DIR}/kvspic-src"
 	BINARY_DIR        "${CMAKE_CURRENT_BINARY_DIR}/kvspic-build"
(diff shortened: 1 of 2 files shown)


unicornss (approved):
looks good.

CMake/Dependencies/libkvspic-CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ include(ExternalProject)
 # clone repo only
 ExternalProject_Add(libkvspic-download
 	GIT_REPOSITORY    https://github.com/awslabs/amazon-kinesis-video-streams-pic.git
-    	GIT_TAG           76223690da1329d733309e2a52575d6a98a68981
```
disa6302, line 10: Unintended change since there is a PR for it already?

disa6302 (approved):
Rest LGTM.

CMake/Dependencies/libkvspic-CMakeLists.txt
hassanctech: Yeah at the time I put out this PR, the other one hadn't been merged -- I'll update branch before merging.

Issue #169, opened 2021-01-08
[BUG] Stream state machine exits when 0x15000011


Nomidia:
**Describe the bug**
When there is something wrong with the state machine, continuousRetryStreamErrorReportHandler will be called. It will check whether the statusCode meets the reconnection conditions by IS_RETRIABLE_ERROR. If the check passes, it will reset the stream.
In some scenarios, when state machine returns 0x15000011, actually it does not happend.

Scenario 1：
1. The describStream returns 52000011
2. ResetStream is called
3. In client state machine, expiration is reached, attempt to get a new expiration time, but timed out. return 0x15000011
4. Check IS_RETRIABLE_ERROR failed, exit the stream state machine.
https://github.com/awslabs/amazon-kinesis-video-streams-producer-c/blob/e1dd9364c3860f0196cc24a5b815c82076ab8149/src/source/ContinuousRetryStreamCallbacks.c#L296

```
16:35:43  2020-12-29 08:35:43 ERROR   blockingCurlCall(): Curl perform failed for url https://c3qp4tl980s52m.credentials.iot.ap-south-1.amazonaws.com/role-aliases/dev-kvs-access-role-alias/credentials with result Timeout was reached : Resolving timed out after 3587 milliseconds 
16:35:43  2020-12-29 08:35:43 WARN    stepStateMachine(): pStateMachine=d08644,end. retStatus=15000011.
16:35:43  2020-12-29 08:35:43 WARN    executeDescribeStreamState(): ---- Leave retStatus=15000011.
16:35:43  
16:35:44  2020-12-29 08:35:43 WARN    stepStateMachine(): pStateMachine=c8764c,end. retStatus=15000011.
16:35:44  2020-12-29 08:35:43 ERROR   describeStreamResultEvent(): operation returned status code: 0x15000011
16:35:44  2020-12-29 08:35:43 WARN    continuousRetryStreamErrorReportHandler(): Reporting stream error. Errored timecode: 0 Status: 0x15000011
```
Scenario 2:
1. curl perform failed for url https://s-703775db.kinesisvideo.ap-south-1.amazonaws.com/putMedia
2. stepStateMachine in kinesisVideoStreamTerminated
3. return 0x15000011, but it does not get this error code.
4. Skip retry in continuousRetryStreamErrorReportHandler

[...]

PR #173, opened 2021-01-12
Adding recoverability on C producer and common lib reported errors pe…

MushMal:
…r customers request.

Issue #169 

*Description of changes:*


hassanctech (commented):
Some new test are failing on Linux, see AddressSanitizer complaints it looks like there's a read after free:  https://travis-ci.com/github/awslabs/amazon-kinesis-video-streams-producer-c/jobs/470721584#L7896

tst/ProducerContinuousRetryTest.cpp
```diff
@@ -154,8 +154,179 @@ TEST_F(ProducerContinuousRetryTest, test_stream_recover_after_reset_connnection_
...
+        EXPECT_EQ(STATUS_SUCCESS, putKinesisVideoFrame(streamHandle, &mFrame));
+        updateFrame();
+        THREAD_SLEEP(mFrame.duration / 2); // speed up test
+    }
+
+    DLOGD("Stopping the stream with stream handle %" PRIu64, (UINT64) streamHandle);
+    EXPECT_EQ(STATUS_SUCCESS, stopKinesisVideoStream(streamHandle));
+    EXPECT_EQ(STATUS_SUCCESS, freeKinesisVideoStream(&streamHandle));
+    EXPECT_EQ(1, mConnectionStaleFnCount);
+    EXPECT_LT(1, mDescribeStreamFnCount);
```
hassanctech, line 194: Tests failing because sometimes mDescribeStreamFnCount is 1.  Did you mean EXPECT_LE (which is less than or equal)?

MushMal: Yeah, it’s hard to get the timing precise as these are independent threads doing the network and recovery. Will change to le

MushMal:
Seeing multiple network timeouts/errors in the TravisCI runs. Due to the unexpected nature of the failures in the test run with fault injections, some of the tests fail with address violation results. This is OK and it's expected in this scenario. Re-started the tests

hassanctech:
> Seeing multiple network timeouts/errors in the TravisCI runs. Due to the unexpected nature of the failures in the test run with fault injections, some of the tests fail with address violation results. This is OK and it's expected in this scenario. Re-started the tests

Still failing -- for tests that run in CI I think we should avoid tests that can sometimes cause unexpected behavior and cause the Ci to fail, having to restart the CI constantly is a huge waste of time.

MushMal:
> > Seeing multiple network timeouts/errors in the TravisCI runs. Due to the unexpected nature of the failures in the test run with fault injections, some of the tests fail with address violation results. This is OK and it's expected in this scenario. Re-started the tests
> 
> Still failing -- for tests that run in CI I think we should avoid tests that can sometimes cause unexpected behavior and cause the Ci to fail, having to restart the CI constantly is a huge waste of time.

I understand it is. However, the last few runs I am getting timeouts in random places. Must be Travis CI issue or something. I will try to give this one more shot and see what's causing these issues. The major problem is that these are not possible to reproduce locally

src/source/ContinuousRetryStreamCallbacks.h
```diff
@@ -33,9 +33,22 @@ struct __StreamLatencyStateMachine;
 struct __ConnectionStaleStateMachine;
 
 typedef struct __CallbackStateMachine {
+    // Indicator of stream being ready
+    volatile BOOL streamReady;
```
hassanctech, line 37: Will any of these be read/written by different threads?  If so we should make them ATOMIC_BOOL

tst/CallbacksProviderApiTest.cpp
```diff
@@ -257,7 +257,7 @@ TEST_F(CallbacksProviderApiTest, createDefaultCallbacksProvider_variations)
     EXPECT_EQ(NULL, pClientCallbacks);
 }
 
-TEST_F(CallbacksProviderApiTest, createStreamVerifyCallbackChainInteration)
+TEST_F(CallbacksProviderApiTest, createStreamVerifyCallbackChainIntegration)
```
hassanctech, line 260: nice

src/source/ContinuousRetryStreamCallbacks.h
MushMal: These will get flagged by TSAN at some stage. I am thinking of bulk resolving TSAN issues at once. There are many false-positives that TSAN reports - for instance, these will be reported (if they are) but there is no chance of a race as I believe these are lock protected.

At any stage, we will do TSAN passes once we are done with the current issues

hassanctech: Is there a reason to not opt for ATOMIC_BOOL by default just due to the nature of there being multiple threads here?  Unless of course if the atomic unit that is needed for correctness is beyond one variable then I agree we should prefer locking.

hassanctech: I do think though we should investigate each TSAN issue one at a time and not bulk resolve, it's possible some of them are actually severe.

MushMal: Bulk-resolve doesn't mean shutting TSAN with suppressions. I meant go over the codebase and understand every reported instance. We do have a very high noise ratio with the reported issues. For example, in WebRTC, we have up-to 80% of the reported issues being benign but we still provided sometimes quirky fixes to get to as clean of a TSAN run as possible.

For this commit though, I don't want to extend to TSAN false-positives.

hassanctech: OK yeah it looks like it's always being modified under the syncLock

hassanctech (approved):

PR #191, opened 2021-02-03
Remove curl pause to fix burstiness

hassanctech:
*Issue #, if available:*. https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/issues/490

*Description of changes:*

Instead of going with the CURL MULTI interface this change introduces condition variable so the curl read callback will wait until it is signaled that it has more data.  This will prevent the case that the callback is fired and data is not yet available and previous in this case we would pause curl.  Doing this yields significant benefit in frame end to end latency.  I have attached a graph of numbers I saw in a simple test with KVS stream in the same region as me.  

CurlFixLatency

Notes on CURL MULTI
----------------------
(1) The nice way to use this is to call [curl_multi_perform](https://curl.se/libcurl/c/curl_multi_perform.html) and then [curl_multi_poll](https://curl.se/libcurl/c/curl_multi_poll.html) which will block until timeout or it's woken up by [curl_multi_wakeup](https://curl.se/libcurl/c/curl_multi_wakeup.html) -- this feature to be able to wake up the blocked thread was introduced in curl 7.68 which was released early 2020. At this point I prefer not to require clients to be on this especially since we support embedded devices which may have curl as part of their firmware and not be easy for them to upgrade. 
(2) Another way is to use [curl_multi_perform](https://curl.se/libcurl/c/curl_multi_perform.html) with [curl_multi_wait](https://curl.se/libcurl/c/curl_multi_wait.html) -- but there is no way to "wake up" the [curl_multi_wait](https://curl.se/libcurl/c/curl_multi_wait.html), aside from timeout it will only wake up when activity is detected in one of the transfers.  The problem is when the read callback is called its possible we have 0 bytes available at this time if we return with 0 curl will tear down the connection so this isn't that simple.
(3) The last way of course is to manually call poll() and select().  There's really no reason I see to be using this.

See related pages:
https://curl.se/libcurl/c/curl_multi_poll.html
https://curl.se/libcurl/c/curl_multi_wakeup.html
https://curl.se/libcurl/c/curl_multi_wait.html
https://curl.se/libcurl/c/curl_multi_perform.html


src/source/Response.c
```diff
@@ -273,7 +296,7 @@ VOID terminateCurlSession(PCurlResponse pCurlResponse, UINT64 timeout)
             // give curl sometime to terminate gracefully before actually timing it out.
             THREAD_SLEEP(timeout);
             // unpause curl in case curl is paused
-            curl_easy_pause(pCurlResponse->pCurl, CURLPAUSE_SEND_CONT);
+            //curl_easy_pause(pCurlResponse->pCurl, CURLPAUSE_SEND_CONT);
```
disa6302, line 299: Remove comment?

src/source/Response.c
```diff
@@ -382,6 +405,7 @@ STATUS curlCompleteSync(PCurlResponse pCurlResponse)
     CURLcode result;
     PCurlRequest pCurlRequest;
     CHAR headers[MAX_REQUEST_HEADER_COUNT * (MAX_REQUEST_HEADER_STRING_LEN + MAX_REQUEST_HEADER_OUTPUT_DELIMITER)];
+    int still_running = 0;
```
disa6302: Looksl ike we do not use this.

src/source/Response.c
```diff
@@ -445,24 +472,25 @@ STATUS curlCompleteSync(PCurlResponse pCurlResponse)
 STATUS notifyDataAvailable(PCurlResponse pCurlResponse, UINT64 durationAvailable, UINT64 sizeAvailable)
 {
     STATUS retStatus = STATUS_SUCCESS;
-    CURLcode result;
+    CURLMcode result;
+    PCallbacksProvider pCallbacksProvider;
 
     CHK(pCurlResponse != NULL, STATUS_NULL_ARG);
 
+    if (pCurlResponse->pCurlRequest != NULL && pCurlResponse->pCurlRequest->pCurlApiCallbacks != NULL &&
```
disa6302, line 478: If any if these are NULL,`pCallbacksProvider` will be pointing to a random location, correct? Wouldn't this cause problems on LN 491?

src/source/Response.h
```diff
@@ -13,12 +13,11 @@ extern "C" {
 // Setting this timeout to terminate CURL connection
 #define TIMEOUT_AFTER_STREAM_STOPPED (1 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND)
 
+#define TIMEOUT_TEARDOWN_CURL_NO_DATA (60ULL * HUNDREDS_OF_NANOS_IN_A_SECOND)
```
MushMal, line 16: We can use 30 seconds as this is what we have in curl configuration - 30 bytes in 30 seconds. We could use the same definition here

src/source/Response.c
```diff
@@ -93,12 +97,25 @@ STATUS freeCurlResponse(PCurlResponse* ppCurlResponse)
...
+        if (IS_VALID_MUTEX_VALUE(pCurlResponse->lock)) {
+            locked = pCallbacksProvider->clientCallbacks.tryLockMutexFn(pCallbacksProvider->clientCallbacks.customData, pCurlResponse->lock);
+            // if it fail to acquire the lock that means another thread is tearing down the response already.
+            // thus current thread can return early.
+            // TODO: eliminate the possibility of two threads calling freeCurlResponse.
+            CHK(locked, retStatus);
+        }
+        if (IS_VALID_MUTEX_VALUE(pCurlResponse->dataAvailableLock)) {
+            dataAvailableLocked = pCallbacksProvider->clientCallbacks.tryLockMutexFn(pCallbacksProvider->clientCallbacks.customData, pCurlResponse->dataAvailableLock);
+            CHK(dataAvailableLocked, retStatus);
```
MushMal, line 110: Need some comments here. So, if we were unable to lock it then does it mean that another teardown is in progress or it means that there is an ongoing operation? In case of the latter, we won't properly teardown the response object (or is it)

src/source/Response.c
hassanctech: Yes you're right, I initialized them null and check to make sure it's not null before using in new revision

src/source/Response.c
MushMal: Perhaps, here, you could trigger an exit with the condition variable and later, when freeing the lock, you could also free the condition variable. NOTE: If you create the condition variable before the data lock, you won't need to check whether its value is valid here.

Think if you need to signal/broadcast termination and await for the condition variable as you hold the lock (try data lock succeeds)

src/source/Response.c
MushMal: @disa6302 if that's the case then all bets are off - you have a mem corruption

src/source/Response.c
```diff
@@ -445,24 +469,24 @@ STATUS curlCompleteSync(PCurlResponse pCurlResponse)
...
-            pCurlResponse->paused = FALSE;
-            // frequent pause unpause causes curl segfault in offline scenario
-            THREAD_SLEEP(10 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
-            // un-pause curl
-            result = curl_easy_pause(pCurlResponse->pCurl, CURLPAUSE_SEND_CONT);
-            if (result != CURLE_OK) {
-                DLOGW("Failed to un-pause curl with error: %u. Curl object %p", result, pCurlResponse->pCurl);
-            }
+        if (pCurlResponse->pCurl != NULL) {
+            pCallbacksProvider->clientCallbacks.lockMutexFn(pCallbacksProvider->clientCallbacks.customData, pCurlResponse->dataAvailableLock);
```
MushMal, line 488: No need to lock the data lock. You can simply proceed with signaling

src/source/Response.c
```diff
@@ -605,8 +629,7 @@ SIZE_T postWriteCallback(PCHAR pBuffer, SIZE_T size, SIZE_T numItems, PVOID cust
     return dataSize;
 }
 
-SIZE_T postReadCallback(PCHAR pBuffer, SIZE_T size, SIZE_T numItems, PVOID customData)
-{
+SIZE_T postReadCallback(PCHAR pBuffer, SIZE_T size, SIZE_T numItems, PVOID customData) {
```
MushMal, line 633: revert?

src/source/Response.c
```diff
@@ -625,26 +648,42 @@ SIZE_T postReadCallback(PCHAR pBuffer, SIZE_T size, SIZE_T numItems, PVOID custo
...
     uploadHandle = pCurlResponse->pCurlRequest->uploadHandle;
 
-    if (pCurlResponse->paused) {
-        bytesWritten = CURL_READFUNC_PAUSE;
-        CHK(FALSE, retStatus);
-    }
-
     if (pCurlResponse->endOfStream || ATOMIC_LOAD_BOOL(&pCurlRequest->requestInfo.terminating)) {
-        DLOGI("Closing connection for upload stream handle: %" PRIu64, uploadHandle);
+        DLOGI("Closing connection for upload stream handle: %"
```
MushMal, line 653: DLOGD is more appropriate here IMHO

src/source/Response.c
```diff
@@ -625,26 +648,42 @@ SIZE_T postReadCallback(PCHAR pBuffer, SIZE_T size, SIZE_T numItems, PVOID custo
...
-            pCurlApiCallbacks->curlReadCallbackHookFn(pCurlResponse, uploadHandle, (PBYTE) pBuffer, (UINT32) bufferSize, &retrievedSize, retStatus);
+                pCurlApiCallbacks->curlReadCallbackHookFn(pCurlResponse, uploadHandle, (PBYTE) pBuffer,
+                                                          (UINT32) bufferSize, &retrievedSize, retStatus);
     }
 
     bytesWritten = (SIZE_T) retrievedSize;
 
+    // If conditions are met under which we would have previously called curl pause -- instead wait on CVAR
+    // SIGNAL is in notifyDataAvailable
+    if (bytesWritten == 0 && (retStatus == STATUS_SUCCESS || retStatus == STATUS_NO_MORE_DATA_AVAILABLE || retStatus == STATUS_AWAITING_PERSISTED_ACK)) {
```
MushMal, line 671: I would suggest doing a quick check first and then do if (bytesWritten == 0)

CHK(etStatus == STATUS_SUCCESS || retStatus == STATUS_NO_MORE_DATA_AVAILABLE || retStatus == STATUS_AWAITING_PERSISTED_ACK, retStatus);

src/source/Response.c
```diff
@@ -625,26 +648,42 @@ SIZE_T postReadCallback(PCHAR pBuffer, SIZE_T size, SIZE_T numItems, PVOID custo
...
+        pCurlApiCallbacks->pCallbacksProvider->clientCallbacks.lockMutexFn(pCurlApiCallbacks->pCallbacksProvider->clientCallbacks.customData, pCurlResponse->dataAvailableLock);
+        CVAR_WAIT(pCurlResponse->dataAvailableCvar, pCurlResponse->dataAvailableLock, TIMEOUT_TEARDOWN_CURL_NO_DATA);
```
MushMal, line 673: You will need the following construct here

lock mutex

while (condition has not been met) {
    anotherRetStatus = CVAR_WAIT();
    if (anotherRetStatus == timeout) {
        // Handle the timeout case
    }

   terminate the outer loop
}

unlock the lock


// The outer loop is needed to combat OS specific spurious wakeups

src/source/Response.c
```diff
@@ -625,26 +648,42 @@ SIZE_T postReadCallback(PCHAR pBuffer, SIZE_T size, SIZE_T numItems, PVOID custo
...
+        pCurlApiCallbacks->pCallbacksProvider->clientCallbacks.unlockMutexFn(pCurlApiCallbacks->pCallbacksProvider->clientCallbacks.customData, pCurlResponse->dataAvailableLock);
+
+        retStatus = getKinesisVideoStreamData(pCurlResponse->pCurlRequest->streamHandle, uploadHandle, (PBYTE) pBuffer,
```
MushMal, line 676: Perhaps, the approach should be something like this

mutex lock
while (bytesWritten == 0) {
    awaitStatus = CVAR_WAIT();
    // Deal with a timeout
    if (awaitStatus == TIMEOUT) {
        retStatus = terminate/abort; ???
    }

    // Proceed with call to get data
    retStatus = getStreamData();
    Proceed with handling of the bytes - switch case with possible abort on bad retStatus
}
mutex unlock

src/source/Response.c
```diff
@@ -51,6 +50,11 @@ STATUS createCurlResponse(PCurlRequest pCurlRequest, PCurlResponse* ppCurlRespon
     pCurlResponse->lock = pCallbacksProvider->clientCallbacks.createMutexFn(pCallbacksProvider->clientCallbacks.customData, TRUE);
     CHK(pCurlResponse->lock != INVALID_MUTEX_VALUE, STATUS_INVALID_OPERATION);
 
+    pCurlResponse->dataAvailableLock = pCallbacksProvider->clientCallbacks.createMutexFn(pCallbacksProvider->clientCallbacks.customData, TRUE);
+    CHK(pCurlResponse->dataAvailableLock != INVALID_MUTEX_VALUE, STATUS_INVALID_OPERATION);
```
MushMal, line 54: Minor: Would recommend STATUS_NOT_ENOUGH_MEMORY instead.

MushMal, line 54: Also minor: use CHK(IS_VALID_MUTEX_VALUE(pCurlResourse->dataAvailableLock));

src/source/Response.c
```diff
@@ -93,12 +97,29 @@ STATUS freeCurlResponse(PCurlResponse* ppCurlResponse)
...
+    if (pCallbacksProvider != NULL) {
+        if (IS_VALID_MUTEX_VALUE(pCurlResponse->lock)) {
+            locked = pCallbacksProvider->clientCallbacks.tryLockMutexFn(pCallbacksProvider->clientCallbacks.customData, pCurlResponse->lock);
+            // if it fail to acquire the lock that means another thread is tearing down the response already.
+            // thus current thread can return early.
+            // TODO: eliminate the possibility of two threads calling freeCurlResponse.
+            CHK(locked, retStatus);
+        }
+        if (IS_VALID_MUTEX_VALUE(pCurlResponse->dataAvailableLock)) {
+            dataAvailableLocked = pCallbacksProvider->clientCallbacks.tryLockMutexFn(pCallbacksProvider->clientCallbacks.customData, pCurlResponse->dataAvailableLock);
```
MushMal, line 109: You should not lock the dataAvailableLock and signal as if there is a CVAR awaiting, it won't be able to re-acquire the lock to "wake up".

src/source/Response.c
```diff
@@ -93,12 +97,29 @@ STATUS freeCurlResponse(PCurlResponse* ppCurlResponse)
...
+            CHK(dataAvailableLocked, retStatus);
+
+            if (dataAvailableLocked) {
```
MushMal, line 112: If the lock was locked then the entity being signaled is not actively awaiting on the CVAR.

src/source/Response.c
```diff
@@ -435,6 +461,8 @@ STATUS curlCompleteSync(PCurlResponse pCurlResponse)
               pCurlResponse->callInfo.responseData, url, headers);
     }
 
+
```
MushMal, line 464: Extra \n

src/source/Response.c
MushMal: Yeah, definitely don't need to lock - in this case, the CVAR will not be able to re-acquire the lock until you release it.

src/source/Response.c
```diff
@@ -625,25 +655,76 @@ SIZE_T postReadCallback(PCHAR pBuffer, SIZE_T size, SIZE_T numItems, PVOID custo
...
-        CHK(FALSE, retStatus);
+    pCurlApiCallbacks->pCallbacksProvider->clientCallbacks.lockMutexFn(
+            pCurlApiCallbacks->pCallbacksProvider->clientCallbacks.customData,
+            pCurlResponse->dataAvailableLock);
+
+    while (bytesWritten == 0 && !ATOMIC_LOAD_BOOL(&pCurlResponse->dataAvailable)) {
+        awaitStatus = CVAR_WAIT(pCurlResponse->dataAvailableCvar, pCurlResponse->dataAvailableLock,
+                                DEFAULT_LOW_SPEED_TIME_LIMIT);
+        if (awaitStatus == STATUS_OPERATION_TIMED_OUT) {
+            DLOGW("Time out...");
```
MushMal, line 672: Perhaps add something more informative like "Awaiting for data to be sent timed out".

This said, perhaps we don't even need any of this. Think if we need to await infinitely. There will be a curl timeout that will signal the exit anyway. You just need to understand that the response is being torn down.

src/source/Response.c
```diff
@@ -625,25 +655,76 @@ SIZE_T postReadCallback(PCHAR pBuffer, SIZE_T size, SIZE_T numItems, PVOID custo
...
+            retStatus = awaitStatus;
+            break;
```
MushMal, line 674: Try to structure your logic to avoid break and continue clauses. I always recommend avoiding these keywords

src/source/Response.c
```diff
@@ -625,25 +655,76 @@ SIZE_T postReadCallback(PCHAR pBuffer, SIZE_T size, SIZE_T numItems, PVOID custo
...
+            retStatus = getKinesisVideoStreamData(pCurlResponse->pCurlRequest->streamHandle, uploadHandle, (PBYTE) pBuffer,
+                                                  (UINT32) bufferSize, &retrievedSize);
+
+            if (pCurlApiCallbacks->curlReadCallbackHookFn != NULL) {
+                retStatus =
+                        pCurlApiCallbacks->curlReadCallbackHookFn(pCurlResponse, uploadHandle, (PBYTE) pBuffer,
+                                                                  (UINT32) bufferSize, &retrievedSize, retStatus);
+            }
+
+            DLOGW("Real wakeup.... bytes Written should NOT be zero");
```
MushMal, line 689: It can be in case of the EOS.

src/source/Response.c
```diff
@@ -625,25 +655,76 @@ SIZE_T postReadCallback(PCHAR pBuffer, SIZE_T size, SIZE_T numItems, PVOID custo
...
+            bytesWritten = (SIZE_T) retrievedSize;
+
+            if (bytesWritten > 0) {
+                ATOMIC_STORE_BOOL(&pCurlResponse->dataAvailable, FALSE);
+            }
+
+            if(retStatus == STATUS_AWAITING_PERSISTED_ACK) {
```
MushMal, line 696: I would strongly recommend re-structuring this loop to avoid break and continue.

src/source/Response.c
```diff
@@ -625,25 +655,76 @@ SIZE_T postReadCallback(PCHAR pBuffer, SIZE_T size, SIZE_T numItems, PVOID custo
...
+        }
     }
 
-    retStatus =
-        getKinesisVideoStreamData(pCurlResponse->pCurlRequest->streamHandle, uploadHandle, (PBYTE) pBuffer, (UINT32) bufferSize, &retrievedSize);
 
-    if (pCurlApiCallbacks->curlReadCallbackHookFn != NULL) {
-        retStatus =
-            pCurlApiCallbacks->curlReadCallbackHookFn(pCurlResponse, uploadHandle, (PBYTE) pBuffer, (UINT32) bufferSize, &retrievedSize, retStatus);
+    if (ATOMIC_LOAD_BOOL(&pCurlResponse->dataAvailable)) {
```
MushMal, line 705: Why do you need to call getKinesisVideoStreamData again? As I recommended, you could handle everything in a single while loop

src/source/Response.c
hassanctech: Ok, we have about a half dozen places where we do this, so I can change all of them.

MushMal: I don't think so. This is really a minor question.

src/source/Response.c
hassanctech: Not sure I follow.  https://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_cond_wait.html

The wait is happening in postReadCallback, under the lock, but when we get to the CVAR_WAIT we will actually release the mutex and block, so yes it is possible for it to be actively waiting on the CVAR once I acquire this lock.

src/source/Response.c
hassanctech: Ok I see the reason it won't is because we will have deleted the mutex before the other thread we signaled gets a chance to acquire it, so it cannot acquire it.

src/source/Response.c
hassanctech: What will happen is it will wake up, realize it can't acquire the lock and just block forever or timeout, I see.  That was only there for testing purposes in the middle of some changes right now.

src/source/Response.c
hassanctech: I basically need to 1.  signal cvar and 2. set the data available boolean to TRUE atomically, hence the lock.  The CVAR is able to re-acquire the lock, I mean this does work, the lock is released immediately after the signal so it should be fine.

hassanctech:
> @hassanctech any update on this work? I had it working initially but have been having issues recently I think due to incompatibilities with the latest C++ sdk. It does not upload any media.

@nickplee Yes this approach fundamentally doesn't work and I've paused (no pun intended lol) worked on this for a bit it looks like we will need to use curl MULTI because we just cannot block in the post read callback. The correct solution will require ensuring the post read callback is ONLY invoked when there is data because if you end up there and you have no data then you really only have two options pause or abort (i.e. tear down curl) neither of which we want to do.

The complication currently arises because in the existing flow we call notify data available for 2 cases 1. When there is data and 2. When awaiting for ACK (no data), the latter case can no longer cause this callback to be invoked or else we'll block.  If we block in the read call back then the post write callback is never invoked (because it's a single thread, CURL doesn't allow you to make that part multi threaded) the post write callback is where the ACK processing happens we we absolutely cannot block that thread.

A solution for this is in the works but definitely do not use what's in this PR as its not usable as it stands and I'm going to convert this PR to a draft.

rejeangroleau:
@nickplee I have been watching this issue/PR for a while and it looks like you found a valid way of adressing this issue, what would be the best way for us to try it out? Would you be kind enough to share your changes with us? Either in another branch or a fork, or even patches, anything would work for us (I don't know your relation with this project, sorry). Thanks!

hassanctech:
> @nickplee I have been watching this issue/PR for a while and it looks like you found a valid way of adressing this issue, what would be the best way for us to try it out? Would you be kind enough to share your changes with us? Either in another branch or a fork, or even patches, anything would work for us (I don't know your relation with this project, sorry). Thanks!

As it stands it actually does not work because my approach was blocking in the read callback which means the write callback never gets invoked and that's where we do ACK processing.  I am working on a better solution for this now and hope to be actively updating this PR.

hassanctech:
> @nickplee I have been watching this issue/PR for a while and it looks like you found a valid way of adressing this issue, what would be the best way for us to try it out? Would you be kind enough to share your changes with us? Either in another branch or a fork, or even patches, anything would work for us (I don't know your relation with this project, sorry). Thanks!

@rejeangroleau Can you please try this: https://github.com/awslabs/amazon-kinesis-video-streams-producer-c/pull/204

PR #199, opened 2021-02-23
Adding ability to create an auth callbacks based on pre-existing cred…

MushMal:
…ential provider

*Issue #, if available:*

The extra API introduced here will be very useful for applications that integrate both KVS producer and WebRTC signaling.

### DIFF
--- a/src/include/com/amazonaws/kinesis/video/cproducer/Include.h
+++ b/src/include/com/amazonaws/kinesis/video/cproducer/Include.h
@@ -520,7 +520,7 @@ PUBLIC_API STATUS createOfflineAudioVideoStreamInfoProvider(PCHAR, UINT64, UINT6
  */
 PUBLIC_API STATUS setStreamInfoBasedOnStorageSize(UINT32, UINT64, UINT32, PStreamInfo);
 
-/*
+/**
  * Frees the StreamInfo provider object.
  *
  * @param[in,out] PStreamInfo* StreamInfo provider object to be freed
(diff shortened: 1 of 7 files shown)


hassanctech (approved):

src/include/com/amazonaws/kinesis/video/cproducer/Include.h
```diff
@@ -650,6 +650,33 @@ PUBLIC_API STATUS createStaticAuthCallbacks(PClientCallbacks, PCHAR, PCHAR, PCHA
...
+ *
+ * NOTE: The original Credential Provider object will not be released on free
+ *
+ * @param[in] PCallbacksProvider Pointer to callback provider
+ * @param[in] PAwsCredentialProvider Credential provider object
+ * @param[in,out] PAuthCallbacks* Pointer to pointer to AuthCallback struct
+ *
+ * @return STATUS status of operation
+ */
+PUBLIC_API STATUS createCredentialProviderAuthCallbacks(PClientCallbacks, PAwsCredentialProvider, PAuthCallbacks*);
```
disa6302, line 667: Do we want to rename this to be more specific for the use case? I am okay with this as well though.

disa6302 (approved):

src/include/com/amazonaws/kinesis/video/cproducer/Include.h
MushMal: The naming reflects the fact that we are supplying a CredentialProvider object which is similar to other overloads

PR #224, opened 2021-07-06
Check system for kvspic before doing git checkout

spotaws:
*Description of changes:*
Currently, the CMakeLists.txt assumes that you do not have any of the kvspic libraries and does a git checkout of https://github.com/awslabs/amazon-kinesis-video-streams-pic in order to meet the required dependency. This makes producer-c much harder to package (e.g. RPM/DEB) as most distributions do not permit network access and this forced git call will fail.

This patch checks for the kvspic library dependency that producer-c cares about (libkvspicUtils) and uses it if found. A more specific check could be done here to ensure that the found library has the necessary functionality, but to be honest, it would be easier to use a proper versioning scheme for kvspic and just check for supported version.

If the kvspic library is not found, it falls back to doing the git checkout logic.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,7 +103,6 @@ if(KVSPIC_FOUND)
   link_directories(${KVSPIC_LIBRARY_DIRS})
 else()
   ############# fetch repos that we need do add_subdirectory ############
-
   # repos that we will build using add_subdirectory will be stored in this path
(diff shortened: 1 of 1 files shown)


disa6302 (approved):

disa6302:
Merging this since unit tests fail due to AWS credential issue with non-maintainer PRs. The changes look good.

PR #234, opened 2021-08-18
add support for proxy

shirogaku:
*Issue #, if available:*

*Description of changes:*
Added support for proxy using environment variable.

### DIFF
--- a/src/source/Common/Curl/CurlCall.c
+++ b/src/source/Common/Curl/CurlCall.c
@@ -55,6 +55,19 @@ STATUS blockingCurlCall(PRequestInfo pRequestInfo, PCallInfo pCallInfo)
         curl_easy_setopt(curl, CURLOPT_SSLVERSION, CURL_SSLVERSION_TLSv1_2);
     }
 
+    // setup proxy using environment variable
+    if(getenv("CURLOPT_PROXY") != NULL) {
+        curl_easy_setopt(curl, CURLOPT_PROXY, getenv("CURLOPT_PROXY"));
+    }
+
+    if(getenv("CURLOPT_PROXYUSERNAME") != NULL) {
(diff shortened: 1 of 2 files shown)


src/source/Common/Curl/CurlCall.c
```diff
@@ -55,6 +55,19 @@ STATUS blockingCurlCall(PRequestInfo pRequestInfo, PCallInfo pCallInfo)
...
+    }
+
+    if(getenv("CURLOPT_PROXYUSERNAME") != NULL) {
+        curl_easy_setopt(curl, CURLOPT_PROXYUSERNAME, getenv("CURLOPT_PROXYUSERNAME"));
+    }
+
+    if(getenv("CURLOPT_PROXYPASSWORD") != NULL) {
+        curl_easy_setopt(curl, CURLOPT_PROXYPASSWORD, getenv("CURLOPT_PROXYPASSWORD"));
+    }
+
```
hassanctech, line 70: Let's not hard code these, how about adding #define's here:

https://github.com/awslabs/amazon-kinesis-video-streams-producer-c/blob/master/src/source/Common/Curl/CurlCall.h#L13

MushMal (commented):
Overall, it's a good idea to be able to set the proxy values. The implementation, though, requires a little more generality. Here are couple of suggestions:

* Use GETENV macro instead of getenv() API.
* Most importantly, enhance the RequestInfo structure with the appropriate fields to be passed into the core APIs instead of the APIs reaching out for the information by calling GETENV. This is not only for the layer violation but also for finer control and testability. Your top level application then could read the values using GETENV for example or read input params and set the RequestInfo structure members. NOTE: you need to have a way of identifying in the inner code that the members have been set to skip curl_easy_setopt-ing NULLs. You could use a boolean or perhaps the value of NULL might be an indicator that the proxy information is not set. 
* Do the plumbing to expose the new RequestInfo properly (make sure in every location the structure is NULL-ed first which it should)

PR #247, opened 2021-09-21
Adding high level retries in continuous retry callback chain and a macro to enable/disable the retries

kateyanurag:
*Issue #, if available:*

*Description of changes:*

Adding high level retries in continuous retry callback chain for the following
(1) Error handler
(2) Latency pressure handler
(3) Connection stale handler

Add a macro to enable/disable exponential retries in continuous callbacks


disa6302:
Stale. Closing

PR #256, opened 2021-11-02
Added check for string length of convertTimestampEpoch and associated…

jdelapla:
… error code

*Issue #, if available:*
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/1270


src/source/Common/Util.c
```diff
@@ -18,6 +18,7 @@ STATUS convertTimestampToEpoch(PCHAR expirationTimestampStr, UINT64 nowTime, PUI
     MEMSET(&ioTExpiration, 0x00, SIZEOF(struct tm));
 
     // iot expiration timestamp format "YYYY-MM-DDTHH:mm:SSZ"
+    CHK(STRLEN(expirationTimestampStr) != STRLEN("YYYY-MM-DDTHH:mm:SSZ"), STATUS_TIMESTAMP_STRING_UNRECOGNIZED_FORMAT);
```
MushMal, line 21: Recommend using STRNLEN for expirationTimestampStr and an ARRAY_SIZE for the pre-defined string. The format should be in the private header

MushMal: Also, perhaps rename the error message name to reflect the length name check

disa6302: Can we define this format: "YYYY-MM-DDTHH:mm:SSZ" instead of using the string as is?

disa6302 (dismissed):

PR #276, opened 2022-02-15
Build system

disa6302:
*Issue #, if available:*

*Description of changes:*
- Initial build system set up to include AWS CRT. 
- Build works on Mac but linking fails on Ubuntu when any aws-c-common related functions are used in the files. At the moment, only Mac builds are working.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@ option(USE_MBEDTLS "Use mbedtls as crypto library" OFF)
 option(BUILD_COMMON_LWS "Whether or not to build ProducerC libwebsockets common library" OFF)
 option(BUILD_COMMON_CURL "Whether or not to build ProducerC curl common library" ON)
 option(BUILD_OPENSSL_PLATFORM "If buildng OpenSSL what is the target platform" OFF)
-option(BUILD_AWS_CRT "Build AWS CRT" ON)
+option(BUILD_AWS_CRT "Build AWS CRT" ON) #Note this does not work for linux platform. Build tested on MacOS only
 
 # Developer Flags
 option(BUILD_TEST "Build the testing tree." OFF)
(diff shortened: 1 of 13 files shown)


CMakeLists.txt
```diff
@@ -98,6 +112,22 @@ endif()
...
+        set(CFLIBRARY "-framework CoreFoundation")
+    endif()
+    set(AWS_C_COMMON_LIBRARY "aws-c-common")
+    set(AWS_LC_LIBRARY "AWSLC")
+    set(AWS_S2N_TLS_LIBRARY "s2n")
+    set(AWS_C_CAL_LIBRARY "aws-c-cal")
+    set(AWS_C_IO_LIBRARY "aws-c-io")
+    set(AWS_C_COMPRESSION_LIBRARY "aws-c-compression")
+    set(AWS_C_HTTP_LIBRARY "aws-c-http")
+    set(AWS_SDK_UTILS_LIBRARY "aws-c-sdkutils")
```
hassanctech, line 127: To be consistent with the others can we name this variable `AWS_C_SDK_UTILS_LIBRARY`

CMakeLists.txt
```diff
@@ -98,6 +112,22 @@ endif()
 
 find_package(PkgConfig REQUIRED)
 
+############# Set AWS CRT based variables #############
+if(BUILD_AWS_CRT)
+    if(APPLE)
+        set(CFLIBRARY "-framework CoreFoundation")
```
hassanctech, line 118: Since this is only set for Apple could we unset it right before this just in case it accidentally gets set somehow so later when we call target link we can't accidentally have something which is non-empty for non-Apple., not sure if that is a real concern.

hassanctech (approved):
great work!!

CMakeLists.txt
disa6302: Done.

CMakeLists.txt
disa6302: Its not a problem. Tested it out. But, set it to empty anyways in the new commit.

Issue #319, opened 2022-08-01
[QUESTION] Any snippets or examples using c producer libs curl/sigv4/credentials functions to upload an image to s3?


mjdietzx:
I am working on an IoT device that streams video to KVS (why we use c producer lib), as well as uploading images to s3. C Producer lib has been great, it is working very well. `createDefaultCallbacksProviderWithIotCertificate` is a really nice helper function that makes authenticating simple. My device has its IoT core certificates that have permissions attached to stream to KVS as well as upload photos to s3, and this "just works"

When it comes to uploading photos to s3 it has been such a pain. I can't find any good example of doing this with curl due to sigv4 signing trickiness and likely some s3 gotchas that I am getting caught up on. I would hate to bring in another big sdk like https://github.com/aws/aws-iot-device-sdk-embedded-C, especially when I am on embedded linux and have curl and this lib which seems to be more than enough to do what I need to do (upload an image to s3).

Looking through the functions in this lib, it seems it wouldn't be too hard to upload an image to s3. Can anyone help me out by providing a code snippet or some pseudo-code on how to do this? I would love to do it with the device cert/key, but I could also provide an access key/security token/session token.

Thanks!

MushMal:
Apparently, uploading images (thumbnails for example) is quite a common thing. I would hope that the AWS folks would have some suggestions for this. While I am not very familiar with the SigV4 calculation for S3 puts, those should be pretty straightforward once you iron out the issues with the way the signature is calculated. For KVS, the signature calculation routine as well as the HTTP request furnishing is done in a targeted fashion in the C producer SDK - you might use large parts of IoT credential provider as well as the HTTP routines but you might need to do some adaptations.

Incidentally, architecturally, try to explore ways to do this in the backend via lambda or something. For example, if you have some backend processing on the incoming stream which also decodes the stream then you could just do S3 push of the images from the backend. Another approach would be to have a separate stream of images only which then on the backend can be stored in KVS and retrieved on demand. You could also push those images to S3 if needed. Sorry, not much help here I feel.

mjdietzx:
Thanks for the response @MushMal . 

Uploading images is so common. It is a shame it is so complicated. I am yet to find a good way to upload an image to an s3 bucket while authenticated using `curl`. It is crazy to me..

ie you would think it would be straightforward to get something like: 
```bash
[...]

PR #323, opened 2022-09-07
Make AWS service configurable for generating presigned urls (eg s3)

mjdietzx:
Resolves the question in https://github.com/awslabs/amazon-kinesis-video-streams-producer-c/issues/319

This PR enables users of c producer lib to generate presigned urls (in our case specifically for `s3`, I'm not sure if there are other AWS services this is useful for) without brining in additional dependencies. c producer lib already does all the heavy lifting, these tiny changes that make it more configurable are all that's needed to make this possible!

I did this in the most minimal, non-invasive way possible, taking care not to break any existing apis or change behavior for existing users. This is by no means an ideal final solution / api for presigning urls, I think the team/maintainers here may want to think if a well-tailored API for this functionality is worth adding. It is likely very minimal changes, just some thought is required for the interfaces/functions you present.

Here is a code snippet of how we use c producer lib to generate presigned s3 urls, and then just use curl to upload an image alongside our KVS stream!

```c
STATUS get_signed_url(PCHAR url, PCHAR presigned)
{
    STATUS retStatus = STATUS_SUCCESS;
    PRequestInfo pRequestInfo = NULL;

    CHK_STATUS(createRequestInfo(url, NULL, kvs_region, NULL, NULL, NULL, SSL_CERTIFICATE_TYPE_NOT_SPECIFIED, "curl/7.79.1",
                                 SERVICE_CALL_INFINITE_TIMEOUT, SERVICE_CALL_INFINITE_TIMEOUT, DEFAULT_LOW_SPEED_LIMIT, DEFAULT_LOW_SPEED_TIME_LIMIT,
                                 pAwsCredentials, &pRequestInfo));

    STRNCPY(pRequestInfo->service, "s3", MAX_SERVICE_NAME_LEN);
    pRequestInfo->verb = HTTP_REQUEST_VERB_PUT;
    CHK_STATUS(removeRequestHeader(pRequestInfo, "user-agent"));

    CHK_STATUS(signAwsRequestInfoQueryParam(pRequestInfo));
    strcpy(presigned, pRequestInfo->url);

CleanUp:

    if (STATUS_FAILED(retStatus))
    {
        defaultLogPrint(LOG_LEVEL_ERROR, "", "get_signed_url failed with status %d\n", retStatus);
    }

    if (pRequestInfo != NULL)
    {
        freeRequestInfo(&pRequestInfo);
    }

    return retStatus;
}
```
Now just take `presigned` and It is as easy as: `curl "<presigned_url>" -H "content-type: image/jpeg" --upload-file /path/to/image.jpg` to upload an image to s3!

---

I am convinced 18071d7d01fd504c34c5ef9011311ef0ae7b56ee fixes a bug. If not please lmk, I think there are other ways to workaround this but from all my tests it really seemed this is the way to go.

---

_In general, I think `service` should be configurable, instead of always just being `kinesisvideo`. Some of the curl requests are to the AWS `iot` service, and in those cases it is incorrectly set to `kinesisvideo` by default, but still works. Still, it doesn't seem ideal. I am not touching any of that code in this PR, but it is somewhat related and may be considered as part of a later related PR/improvements._


src/include/com/amazonaws/kinesis/video/common/Include.h
```diff
@@ -102,6 +102,11 @@ extern "C" {
  */
 #define MAX_REGION_NAME_LEN 128
 
+/**
+ * Maximum allowed service name length
+ */
+#define MAX_SERVICE_NAME_LEN 32
```
MushMal, line 108: Perhaps someone from AWS can comment on the max length?

src/include/com/amazonaws/kinesis/video/common/Include.h
```diff
@@ -482,6 +487,7 @@ struct __RequestInfo {
                                               //!< NOTE: The body will follow the main struct
     UINT32 bodySize;                          //!< Size of the body in bytes
     CHAR url[MAX_URI_CHAR_LEN + 1];           //!< The URL for the request
+    PCHAR service;                            //!< The AWS service for the request
```
MushMal, line 490: Strongly recommend having a char array of size MAX_SERVICE_NAME_LEN and copy into it with zero termination and checking for overflow. This is important for multiple reasons, including the encapsulation and deep isolation which can make the structure easy to pass around boundaries and execution units (including different threads)

src/source/Common/AwsV4Signer.c
```diff
@@ -73,7 +73,7 @@ STATUS generateAwsSigV4Signature(PRequestInfo pRequestInfo, PCHAR dateTimeStr, B
     hmacSize = SIZEOF(hmac);
     CHK_STATUS(generateRequestHmac((PBYTE) pScratchBuf, curSize, (PBYTE) dateTimeStr, SIGNATURE_DATE_STRING_LEN * SIZEOF(CHAR), hmac, &hmacSize));
     CHK_STATUS(generateRequestHmac(hmac, hmacSize, (PBYTE) pRequestInfo->region, (UINT32) STRLEN(pRequestInfo->region), hmac, &hmacSize));
-    CHK_STATUS(generateRequestHmac(hmac, hmacSize, (PBYTE) KINESIS_VIDEO_SERVICE_NAME, (UINT32) STRLEN(KINESIS_VIDEO_SERVICE_NAME), hmac, &hmacSize));
+    CHK_STATUS(generateRequestHmac(hmac, hmacSize, (PBYTE) pRequestInfo->service, (UINT32) STRLEN(pRequestInfo->service), hmac, &hmacSize));
```
MushMal, line 76: STRLEN is a very dangerous API. In the previous case, it should not have been used but instead used ARRAY_SIZE. Either way, it was hard-coded value which was kind of OK. In this case, the size should be calculated by calling STRNLEN - with max size of MAX_SERVICE_NAME_LEN. STRNLEN should also be used in the lines above and below.

src/source/Common/AwsV4Signer.c
```diff
@@ -117,6 +117,12 @@ STATUS signAwsRequestInfo(PRequestInfo pRequestInfo)
 
     CHK(pRequestInfo != NULL && pRequestInfo->pAwsCredentials != NULL, STATUS_NULL_ARG);
 
+    // signAwsRequestInfo is a public api function, if service is not specified default to "kinesisvideo" so no breaking changes are introduced to the api
+    if (pRequestInfo->service == NULL)
```
MushMal, line 121: As recommended, array should be used instead of a raw pointer. In that case, the check here should be something like

if (pRequestInfo->service[0] == L'\0') { //set to kinesisvideo"} 

Ideally, this logic should be applied at the caller layer rather than here. This should simply check for it not being NULL and fail if it's null

src/source/Common/AwsV4Signer.c
```diff
@@ -167,6 +173,12 @@ STATUS signAwsRequestInfoQueryParam(PRequestInfo pRequestInfo)
 
     CHK(pRequestInfo != NULL && pRequestInfo->pAwsCredentials != NULL, STATUS_NULL_ARG);
 
+    // signAwsRequestInfoQueryParam is a public api function, if service is not specified default to "kinesisvideo" so no breaking changes are introduced to the api
```
MushMal, line 176: Same comment

src/source/Common/AwsV4Signer.c
```diff
@@ -498,8 +510,14 @@ STATUS generateCanonicalRequestString(PRequestInfo pRequestInfo, PCHAR pRequestS
     len = SHA256_DIGEST_LENGTH * 2;
     CHK(curLen + len <= requestLen, STATUS_BUFFER_TOO_SMALL);
     if (pRequestInfo->body == NULL) {
-        // Streaming treats this portion as if the body were empty
-        CHK_STATUS(hexEncodedSha256((PBYTE) EMPTY_STRING, 0, pCurPtr));
+        if (!STRNCMP(pRequestInfo->service, KINESIS_VIDEO_SERVICE_NAME, MAX_SERVICE_NAME_LEN)) {
```
MushMal, line 513: STRNCMP returns an integer so I would recommend using an explicit check the result against being 0

src/source/Common/AwsV4Signer.c
```diff
@@ -498,8 +510,14 @@ STATUS generateCanonicalRequestString(PRequestInfo pRequestInfo, PCHAR pRequestS
     len = SHA256_DIGEST_LENGTH * 2;
     CHK(curLen + len <= requestLen, STATUS_BUFFER_TOO_SMALL);
     if (pRequestInfo->body == NULL) {
-        // Streaming treats this portion as if the body were empty
-        CHK_STATUS(hexEncodedSha256((PBYTE) EMPTY_STRING, 0, pCurPtr));
+        if (!STRNCMP(pRequestInfo->service, KINESIS_VIDEO_SERVICE_NAME, MAX_SERVICE_NAME_LEN)) {
+            // Streaming treats this portion as if the body were empty
+            CHK_STATUS(hexEncodedSha256((PBYTE) EMPTY_STRING, 0, pCurPtr));
+        } else {
+            PCHAR term = "UNSIGNED-PAYLOAD";
```
MushMal, line 517: This should be defined in the internal header. The length in that case should be calculated as ARRAY_SIZE(PREDEFINED_UNSIGNED_PAYLOAD)

src/source/Common/AwsV4Signer.c
```diff
@@ -678,14 +696,14 @@ STATUS generateCredentialScope(PRequestInfo pRequestInfo, PCHAR dateTimeStr, PCH
     CHK(pRequestInfo != NULL && dateTimeStr != NULL && pScopeLen != NULL, STATUS_NULL_ARG);
 
     // Calculate the max string length with a null terminator at the end
-    scopeLen = SIGNATURE_DATE_TIME_STRING_LEN + 1 + MAX_REGION_NAME_LEN + 1 + (UINT32) STRLEN(KINESIS_VIDEO_SERVICE_NAME) + 1 +
+    scopeLen = SIGNATURE_DATE_TIME_STRING_LEN + 1 + MAX_REGION_NAME_LEN + 1 + (UINT32) STRLEN(pRequestInfo->service) + 1 +
```
MushMal, line 699: Should use STRNLEN

src/source/Common/AwsV4Signer.c
```diff
@@ -707,15 +725,15 @@ STATUS generateEncodedCredentials(PRequestInfo pRequestInfo, PCHAR dateTimeStr,
     CHK(pRequestInfo != NULL && dateTimeStr != NULL && pCredsLen != NULL, STATUS_NULL_ARG);
 
     // Calculate the max string length with '/' and a null terminator at the end
-    credsLen = MAX_ACCESS_KEY_LEN + 1 + SIGNATURE_DATE_TIME_STRING_LEN + 1 + MAX_REGION_NAME_LEN + 1 + (UINT32) STRLEN(KINESIS_VIDEO_SERVICE_NAME) +
+    credsLen = MAX_ACCESS_KEY_LEN + 1 + SIGNATURE_DATE_TIME_STRING_LEN + 1 + MAX_REGION_NAME_LEN + 1 + (UINT32) STRLEN(pRequestInfo->service) +
```
MushMal, line 728: Same

MushMal (commented):
I think this is a great idea and a very well worthy commit. I have added a few comments on the implementation side.

mjdietzx:
Thank you for the great review @MushMal . In some cases I tried to stick very closely to how things are currently implemented, but you give great explanations and suggestions in your comments. So I will likely take all your suggestions and get something pushed in the next few days!

src/source/Common/AwsV4Signer.c
mjdietzx: > Ideally, this logic should be applied at the caller layer rather than here. This should simply check for it not being NULL and fail if it's null

Yes, I agree. And I hope some thought and further work goes into this to present a well-designed api for this use-case. I think it is beyond the scope of this PR bc I do not want to break or affect the public api in any way (with the hopes that this PR is extremely minimal/simple and can be merged). I would be happy to help in a followup PR, and definitely review and adopt any PRs that aim to extend this.

src/include/com/amazonaws/kinesis/video/common/Include.h
mjdietzx: I went ahead and made this `128` - there is some precedent as that is what max region len is. I have also never seen an aws service even close to this length, so am confident it is safe

mjdietzx:
Thanks again for the great review @MushMal ! I have made all fixes according to your comments and marked them resolved. I have refrained from additional changes outside the scope of this PR like changing existing `strlen` -> `strnlen` for the sake of keeping the scope of this PR as limited as possible. As well as any api facing changes (although I agree that is useful future work / followup PRs)

mjdietzx:
Noticed that PRs go to `develop` and not `master`. Rebased and forced push to correct this

src/source/Common/AwsV4Signer.c
```diff
@@ -117,6 +117,11 @@ STATUS signAwsRequestInfo(PRequestInfo pRequestInfo)
 
     CHK(pRequestInfo != NULL && pRequestInfo->pAwsCredentials != NULL, STATUS_NULL_ARG);
 
+    // signAwsRequestInfo is a public api function, if service is not specified default to "kinesisvideo" so no breaking changes are introduced to the api
+    if (pRequestInfo->service[0] == L'\0') {
```
MushMal, line 121: A small nit. I would not modify a structure that's passed to this layer by a higher-layer. Instead, I would define a pointer to char like

PCHAR pServiceName = (PCHAR) KINESIS_VIDEO_SERVICE_NAME; 

in the beginning of the function. Then, here, I would do something like

if (pRequestInfo->service[0] != L'\0') {
    pServiceName = pRequestInfo->service;
}

This means that any place where pRequestInfo->service is referenced needs to reference pServiceName.

This is a minor nit though

src/source/Common/AwsV4Signer.c
```diff
@@ -167,6 +172,11 @@ STATUS signAwsRequestInfoQueryParam(PRequestInfo pRequestInfo)
 
     CHK(pRequestInfo != NULL && pRequestInfo->pAwsCredentials != NULL, STATUS_NULL_ARG);
 
+    // signAwsRequestInfoQueryParam is a public api function, if service is not specified default to "kinesisvideo" so no breaking changes are introduced to the api
+    if (pRequestInfo->service[0] == L'\0') {
+        STRNCPY(pRequestInfo->service, KINESIS_VIDEO_SERVICE_NAME, MAX_SERVICE_NAME_LEN);
```
MushMal, line 177: same comment as the one before about mutating the passed-in structure

MushMal (approved):
minor comments

src/source/Common/AwsV4Signer.c
mjdietzx: Hmm, instead of modifying anything I could always do it just in time, ie just replace `pRequestInfo->service` everywhere with:
```c
pRequestInfo->service[0] == L'\0' ? KINESIS_VIDEO_SERVICE_NAME : pRequestInfo->service
```

It seems the way you are suggesting here would be tricky bc this is used by other functions `generateAwsSigV4Signature`, `generateCanonicalRequestString`, `generateCredentialScope`, `generateEncodedCredentials` that pass around `pRequestInfo`

If its a minor nit I'd prefer to leave as-is (if this is acceptable), unless there is a very simple solution

MushMal: I agree, it really is a minor issue and not worth adding more complexity.

PR #325, opened 2022-09-14
make readFileCredentials return an error if credentials are expired

mjdietzx:
`readFileCredentials` is currently a little confusing:

I auth with `createDefaultCallbacksProviderWithFileAuth(CREDENTIALS_FILE_PATH, kvs_region, AMAZON_CA_CERT_PATH, NULL, NULL, &pClientCallbacks);`, and really expected this to fail if the credentials were expired (I cache credentials for up to 12 hours / until expiry in `CREDENTIALS_FILE_PATH`). Basically these underlying functions do all the checks to see if credentials are expired, but allow expired credentials to be loaded (only if `pFileCredentialProvider->pAwsCredentials` is not set yet, once it's set expired credentials will not be loaded - basically `readFileCredentials` is a no-op in that case). That differing behavior is also confusing, and I really had to dig through code to understand it (hence I provide some comments to clarify it a bit).

I think a better api is for `readFileCredentials` to return the `STATUS_IOT_EXPIRATION_OCCURS_IN_PAST` error when expired credentials are attempted to be loaded (for the first time). That way I can just catch the error when I do the top-level `createDefaultCallbacksProviderWithFileAuth(CREDENTIALS_FILE_PATH, kvs_region, AMAZON_CA_CERT_PATH, NULL, NULL, &pClientCallbacks);` and wait the 1-2s for my new/fresh credentials to come in.

Otherwise, it looks like `createDefaultCallbacksProviderWithFileAuth` succeeds, with no indication credentials are expired, and then I'll get an error later, ie: 
```
INFO    describeStreamResultEvent(): Describe stream result event.
INFO    writeHeaderCallback(): RequestId: 087c5f43-e656-4313-8d89-1112ebb1de22
WARN    curlCompleteSync(): HTTP Error 403 : Response: {"message":"The security token included in the request is expired"}
```

This then retries a bunch of times, but even if new credentials are loaded into `CREDENTIALS_FILE_PATH` during this time, they are never re-loaded into the credential provider automatically during the retries. That's a bit of a different issue, it would be nice, but I think this PR is a good/very simple start.

### DIFF
--- a/src/source/Common/FileCredentialProvider.c
+++ b/src/source/Common/FileCredentialProvider.c
@@ -121,9 +121,9 @@ STATUS readFileCredentials(PFileCredentialProvider pFileCredentialProvider)
 
     CHK(pFileCredentialProvider != NULL && pFileCredentialProvider->credentialsFilepath != NULL, STATUS_NULL_ARG);
 
-    // Refresh the credentials by reading from the credentials file if needed
+    // Only refresh the credentials by reading from the credentials file if needed.
+    // If we already have credentials and they are not expiring soon, we return successfully here.
     currentTime = pFileCredentialProvider->getCurrentTimeFn(pFileCredentialProvider->customData);
-
     CHK(pFileCredentialProvider->pAwsCredentials == NULL ||
             currentTime + CREDENTIAL_FILE_READ_GRACE_PERIOD > pFileCredentialProvider->pAwsCredentials->expiration,
         retStatus);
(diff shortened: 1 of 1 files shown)


MushMal (dismissed):
The change itself looks good. Regarding the expiration... I agree that perhaps the right course of action should be to do validation in createAwsCredentials API itself. The change should be put through set of tests to ensure we are not breaking anything in the background refresh and session re-streaming.

mjdietzx:
The failing checks seem to be unrelated to this PR:
```
### Configure AWS Credentials
Run aws-actions/configure-aws-credentials@v1
  with:
    role-duration-seconds: 10800
    audience: sts.amazonaws.com
  env:
    AWS_KVS_LOG_LEVEL: 2
    LDFLAGS: -L/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib
    CPATH: /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/
Error: Input required and not supplied: aws-region
```

MushMal (approved):

jdelapla (approved):

jdelapla:
Thank you for your submission!

PR #378, opened 2023-07-13
Remove an extra newline in the logs

daveisfera:
There's an extra newline in the logs that doesn't need to be there


daveisfera:
Done. I'd recommend changing the default branch to `develop` so that happens automatically in the future

daveisfera:
Never mind. Looks like that switch already and my fork is defaulting to `master` because it's older

sirknightj (approved):

PR #382, opened 2023-07-26
Fix the usage of the samples

daveisfera:
The usage of the samples wasn't updated when the changes were made, so this resolves that


disa6302 (approved):
Thank you!

daveisfera:
I'm guessing it was an accident, but [this commit](https://github.com/awslabs/amazon-kinesis-video-streams-producer-c/commit/f60ae7d8260353e8f9b5e38df11a845da53bc99f#diff-abff77b438faaead1f7f12496286e23c8a3672f7c88f9fcfdd52952b728e13baR213) reverted this fix

PR #391, opened 2023-08-09
Fix deprecated usage of curl

daveisfera:
Fix deprecated usage of `curl` that is warned about when building


hassanctech (approved):
Thank you for this submission!

PR #394, opened 2023-08-10
Add setting of fragment metadata to the sample

daveisfera:
It appears that the fragment metadata is not consistently stored in KVS. Sometimes the data is available when downloading using `get_media` but sometimes it's not. To verify this behavior, I added the ability to set fragment metadata to one of the samples, so this is that change and can be used to verify the functionality


sirknightj:
Merged linked one

PR #408, opened 2023-09-14
Set newly added timeouts to default value

daveisfera:
Timeouts were added but the defaults weren't set, so this fixes that so warnings aren't output needlessly


sirknightj:
Thanks daveisfera!

Also in case anyone is wondering what warning is being referred to, it is this one:
```
[WARN ] [20-09-2023 20:44:25:468.759 GMT] fixupClientInfo(): Connection timeout is invalid...setting to default
[WARN ] [20-09-2023 20:44:25:468.812 GMT] fixupClientInfo(): Completion timeout is invalid...setting to default
```

PR #426, opened 2023-12-20
Update thing name length to be 128 as per AWS docs

disa6302:
*Issue #, if available:*

*Description of changes:*
- Our thing name length is currently set to 256 characters (stream name length). However, as per https://docs.aws.amazon.com/iot/latest/apireference/API_CreateThing.html, it is 128 characters. Since our SDK does not create the thing, the failure would be caught in the out of band thing creation step anyways. However, we need this change to be accurate.


hassanctech (approved):
lgtm

PR #428, opened 2024-01-16
Release build for producer C

disa6302:
*Issue #, if available:*

*Description of changes:*
- The PR ensures the SDK is built in release mode by default. Also fixed unit tests failure caused by `assert` usage which gets optimized out in Release mode. Throwing an error instead to ensure the log level is set to a valid value if set through env.

- Also removed jsm cmake files since the SDK directly includes the jsmn.h file instead of building from source.

Testing:

- Confirmed libraries generated are release libraries through the CMakeCache.txt file
- Testing with CI


CMake/Dependencies/libjsmn-CMakeLists.txt
hassanctech, line 1: Did you mean to delete these?

hassanctech (approved):
lgtm

PR #430, opened 2024-01-17
Handle aws special regions

hassanctech:
*Issue #, if available:*

*Description of changes:*
Adding support for ADC and GovCloud regions (FIPS endpoints) + tests to confirm correct control plane url is created for the supplied region.

### DIFF
--- a/src/source/CurlApiCallbacks.c
+++ b/src/source/CurlApiCallbacks.c
@@ -79,12 +79,23 @@ STATUS createCurlApiCallbacks(PCallbacksProvider pCallbacksProvider, PCHAR regio
 
     // Set the control plane URL
     if (controlPlaneUrl == NULL || controlPlaneUrl[0] == '\0') {
-        // Create a fully qualified URI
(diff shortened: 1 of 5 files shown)


tst/ProducerCallbackProviderApiTest.cpp
```diff
@@ -58,6 +58,55 @@ namespace com { namespace amazonaws { namespace kinesis { namespace video {
 
     }
 
+    TEST_F(ProducerCallbackProviderApiTest, TestCorrectControlPlaneUriForSpecifiedRegion)
```
disa6302, line 61: Nice!

disa6302 (approved):

src/source/CurlApiCallbacks.c
```diff
@@ -79,12 +79,23 @@ STATUS createCurlApiCallbacks(PCallbacksProvider pCallbacksProvider, PCHAR regio
 
     // Set the control plane URL
     if (controlPlaneUrl == NULL || controlPlaneUrl[0] == '\0') {
```
sirknightj, line 81: /nit We can use IS_NULL_OR_EMPTY_STRING macro

PR #431, opened 2024-01-18
run windows tests with debug logging

hassanctech:
*Issue #, if available:*

*Description of changes:*
* CALLOC frame buffer to prevent garbage data
* DescribeStream API response wait until stream definition is memset and populated before printing out log line otherwise on some platforms you will get a lot of garbage (we were seeing that on Windows).


jdelapla (approved):

src/source/CurlApiCallbacks.c
```diff
@@ -1441,6 +1440,10 @@ PVOID describeStreamCurlHandler(PVOID arg)
 
 CleanUp:
 
+    if (responseReceived) {
```
sirknightj, line 1443: Equivalent to `resultLen > 0`?

PR #442, opened 2024-02-29
Mbedtls link

disa6302:
*Issue #, if available:*

*What was changed?*
- Fixed the linking for LWS case to be crypto specific.

*Why was it changed?*
- The crypto linking is different for LWS vs CURL. In LWS case for this library, the library was always being linked to openssl, but it should link based on crypto library set up.

*How was it changed?*
- Linking to `PRODUCER_CRYPTO_LIBRARY` instead of Openssl only.

Before this change: (`libmbedcrypto.so.7 => not found`)

```
ubuntu@ip-172-31-11-82:~/amazon-kinesis-video-streams-producer-c/build$ ldd libkvsCommonLws.so
	linux-vdso.so.1 (0x00007fffcbf08000)
	libwebsockets.so.15 => /home/ubuntu/amazon-kinesis-video-streams-producer-c/open-source/lib/libwebsockets.so.15 (0x00007f32d5b7a000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f32d5800000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f32d5bcf000)
	libmbedtls.so.14 => not found
	libmbedx509.so.1 => not found
	libmbedcrypto.so.7 => not found
```

After this change: (`libmbedcrypto.so.7 => /home/ubuntu/amazon-kinesis-video-streams-producer-c/open-source/lib/libmbedcrypto.so.7 (0x00007f8518325000)`)

```
ubuntu@ip-172-31-11-82:~/amazon-kinesis-video-streams-producer-c/build$ ldd libkvsCommonLws.so
	linux-vdso.so.1 (0x00007ffdfdb56000)
	libmbedcrypto.so.7 => /home/ubuntu/amazon-kinesis-video-streams-producer-c/open-source/lib/libmbedcrypto.so.7 (0x00007f8518325000)
	libwebsockets.so.19 => /home/ubuntu/amazon-kinesis-video-streams-producer-c/open-source/lib/libwebsockets.so.19 (0x00007f85182c6000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f8518000000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f85183b4000)
	libmbedtls.so.14 => not found
	libmbedx509.so.1 => not found
```

- Added CI to only test the build

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -176,6 +176,28 @@ jobs:
           cd build
           ./tst/producer_test
 
+  ubuntu-os-build-lws-mbedtls:
+    runs-on: ubuntu-20.04
+    env:
+      AWS_KVS_LOG_LEVEL: 2
+      CC: gcc
+      CXX: g++
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v3
+      - name: Install dependencies
+        run: |
+          sudo apt clean && sudo apt update
+          sudo sh -c 'echo 0 > /proc/sys/net/ipv6/conf/all/disable_ipv6'
+      - name: Build repository
+        run: |
+          mkdir build && cd build
+          cmake .. -DBUILD_COMMON_LWS=ON -DUSE_OPENSSL=OFF -DUSE_MBEDTLS=ON
+          make
+
   linux-gcc-code-coverage:
     runs-on: ubuntu-20.04
     env:
@@ -268,6 +290,7 @@ jobs:
           cd build
           ulimit -c unlimited -S
           timeout --signal=SIGABRT 150m ./tst/producer_test --gtest_break_on_failure
+
 
   # memory-sanitizer:
   #   runs-on: ubuntu-20.04
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -176,7 +176,7 @@ jobs:
           cd build
           ./tst/producer_test
 
-  ubuntu-os-build-lws-mbedtls:
+  ubuntu-os-gcc-build-lws-mbedtls:
     runs-on: ubuntu-20.04
     env:
       AWS_KVS_LOG_LEVEL: 2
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -289,8 +289,7 @@ if(BUILD_COMMON_LWS)
       set_target_properties(kvsCommonLws PROPERTIES VERSION ${KINESIS_VIDEO_PRODUCER_C_VERSION} SOVERSION ${KINESIS_VIDEO_PRODUCER_C_MAJOR_VERSION})
     endif()
     target_link_libraries(kvsCommonLws
-            ${OPENSSL_CRYPTO_LIBRARY}
-            ${OPENSSL_SSL_LIBRARY}
+            ${PRODUCER_CRYPTO_LIBRARY}
```
niyatim23, line 292: Shouldn't there be an if-else for openssl and mbedtls and ${OPENSSL_SSL_LIBRARY} removed only for mbedtls?

.github/workflows/ci.yml
```diff
@@ -176,6 +176,28 @@ jobs:
           cd build
           ./tst/producer_test
 
+  ubuntu-os-build-lws-mbedtls:
```
jdelapla, line 179: add gcc to name

hassanctech:
Nice!  Is it possible to print the results of `ldd` from before / after to confirm we're not linking to openssl any more, i.e. it's not specified anywhere in the `NEEDED` section of the ELF header?

CMakeLists.txt
disa6302: https://github.com/awslabs/amazon-kinesis-video-streams-producer-c/pull/442/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR263-R270

stefankiesz (approved):

disa6302:
> Nice! Is it possible to print the results of `ldd` from before / after to confirm we're not linking to openssl any more, i.e. it's not specified anywhere in the `NEEDED` section of the ELF header?

Done!

niyatim23 (approved):

stefankiesz (approved):

hassanctech:
> > Nice! Is it possible to print the results of `ldd` from before / after to confirm we're not linking to openssl any more, i.e. it's not specified anywhere in the `NEEDED` section of the ELF header?
> 
> Done!

Looking at the results it would say `not found` just because it can't find it in `LD_LIBRARY_PATH` or standard library path but in this case it looks like previously that library was just not built and put in the lib folder which is where it looks for it in the AFTER case. However what I was expecting to see in the BEFORE case is libssl/libcrypto in the list, since it is not there that would means previously it wasn't linking to openssl.

hassanctech (approved):
lgtm

PR #446, opened 2024-04-11
Increase UT coverage

disa6302:
*Issue #, if available:*

*What was changed?*
- Removed samples and jsmn from coverage report
- Added new unit tests for LWS, Curl, AwsV4Signer and IoT paths
- Fixed typos in the code that were caught while adding new unit tests
- Using lcov instead of gcov to allow for better filters in coverage reports (for our purpose, remove jsmn.h from report).

*Why was it changed?*
- To ensure existing APIs were tested via unit tests
- To report the appropriate coverage provided by the unit tests

*How was it changed?*
- Increased test suite
- Moved to lcov to allow for better filtering in the future.
- Increased coverage from 64.24% to 90.81%

- Ran existing and new unit tests

Future changes:
- Include LWS build on Mac (currently, there are some globals defined in PIC that causes our unit tests to fail when building LWS and CURL at the same time)
- Include IoT tests in Windows (currently, there is an issue on windows where the SDK cannot open the cert file even if it exists, needs dedicated investigation).


.github/workflows/codecov.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Run tests
+        run: |
+          cd build
+          ulimit -c unlimited -S
+          source ../scripts/get-iot-certs.sh ci
+          timeout --signal=SIGABRT 150m ./tst/producer_test --gtest_break_on_failure
+      - name: Code coverage
+        run: |
+          cd build
+          for test_file in $(find cproducer.dir kvsCommonCurl.dir kvsCommonLws.dir -name '*.gcno'); do gcov $test_file; done
```
hassanctech, line 46: Was this `gcov` supposed to be changed to `lcov` as well?

src/source/Common/RequestInfo.c
```diff
@@ -245,7 +245,8 @@ STATUS removeRequestHeader(PRequestInfo pRequestInfo, PCHAR headerName)
 
         if (STRCMPI(pCurrentHeader->pName, headerName) == 0) {
             CHK_STATUS(singleListDeleteNode(pRequestInfo->pRequestHeaders, pCurNode));
-
+            // Free only if found
+            SAFE_MEMFREE(pCurrentHeader);
             // Early return
             CHK(FALSE, retStatus);
         }
```
hassanctech, line 252: Ok this logic is only functionally different in the case that we executed these lines and 

```
CHK_STATUS(singleListGetNodeData(pCurNode, &item));
pCurrentHeader = (PRequestHeader) item;
```

```
STRCMPI(pCurrentHeader->pName, headerName) == 0
```

Is `FALSE` in that case previously we would have called SAFE_MEMFREE, but now we do not.  Do we have a targeted test for this case? 

In other places we do a null check right after this line:
```
pCurrentHeader = (PRequestHeader) item;
```

Do we not need that were as well since we deref `pCurrentHeader` later or as we just assuming it can never be null (perhaps there is validation on every insert and we have mutexes to make sure it cannot be made null by another thread?). If the header is allocated but it didn't match the `STRCMPI` it doesn't get free I'm concerned about a leak in that case because I'm not seeing where that would get freed?

tst/AuthCallbackTest.cpp
```diff
@@ -19,13 +19,16 @@ TEST_F(AuthCallbackTest, ioTExpirationParsing_Returns_Success)
...
+    EXPECT_EQ(STATUS_NULL_ARG, convertTimestampToEpoch(NULL, iotTimeInEpoch, NULL));
+    EXPECT_EQ(STATUS_NULL_ARG, convertTimestampToEpoch(validFormatIotExpirationTimeStamp, iotTimeInEpoch, NULL));
+    EXPECT_EQ(STATUS_SUCCESS, convertTimestampToEpoch(validFormatIotExpirationTimeStamp, iotTimeInEpoch, &expirationTimestampInEpoch));
 
     EXPECT_TRUE(iotTimeInEpoch == expirationTimestampInEpoch / HUNDREDS_OF_NANOS_IN_A_SECOND - 3600);
 
     iotTimeInEpoch = 1548975659;   // iot expiration same as current time
 
-    convertTimestampToEpoch(validFormatIotExpirationTimeStamp, iotTimeInEpoch, &expirationTimestampInEpoch);
+    EXPECT_EQ(STATUS_SUCCESS,convertTimestampToEpoch(validFormatIotExpirationTimeStamp, iotTimeInEpoch, &expirationTimestampInEpoch));
```
hassanctech, line 31: nice!

src/source/Common/IotCredentialProvider.c
```diff
@@ -25,7 +25,7 @@ STATUS createIotCredentialProviderWithTime(PCHAR iotGetCredentialEndpoint, PCHAR
     pIotCredentialProvider->getCurrentTimeFn = (getCurrentTimeFn == NULL) ? commonDefaultGetCurrentTimeFunc : getCurrentTimeFn;
     pIotCredentialProvider->customData = customData;
 
-    CHK(STRNLEN(iotGetCredentialEndpoint, MAX_URI_CHAR_LEN + 1) <= MAX_URI_CHAR_LEN, MAX_URI_CHAR_LEN);
+    CHK(STRNLEN(iotGetCredentialEndpoint, MAX_URI_CHAR_LEN + 1) <= MAX_URI_CHAR_LEN, STATUS_INVALID_ARG);
```
hassanctech, line 28: wow good catch!!

.github/workflows/codecov.yml
disa6302: Good catch. This is not required at all since we are generating with lcov. I had added this when I was debugging issues with lcov for comparison purposes.

src/source/Common/RequestInfo.c
hassanctech: I understand now thank you!

src/source/CallbacksProvider.c
```diff
@@ -230,7 +230,7 @@ STATUS createDefaultCallbacksProviderWithFileAuth(PCHAR credentialsFilePath, PCH
         }
 
         if (pAuthCallbacks != NULL) {
-            freeIotAuthCallbacks(&pAuthCallbacks);
+            freeFileAuthCallbacks(&pAuthCallbacks);
```
hassanctech, line 233: Good catch!

src/source/Response.c
```diff
@@ -282,78 +282,6 @@ VOID terminateCurlSession(PCurlResponse pCurlResponse, UINT64 timeout)
     }
 }
 
-VOID dumpResponseCurlEasyInfo(PCurlResponse pCurlResponse)
```
hassanctech, line 285: This is a diagnostic / debug function I think we should keep it. Is there a reason why we're deleting it (other than it's not currently used and is decreasing our overall code line coverage count -- I don't think those are good reasons to delete it).  Instead if that is the reason can we just add a call to this function hidden behind a flag of some sort? If the function itself is buggy then that's a different situation.

src/source/StreamInfoProvider.c
```diff
@@ -112,7 +112,7 @@ STATUS createVideoTrackInfo(VIDEO_CODEC_ID videoCodecId, PCHAR contentType, PTra
             break;
         default:
             STRCPY(pTrackInfo->codecId, MKV_H264_AVC_CODEC_ID);
-            STRCAT(contentType, MKV_H265_CONTENT_TYPE);
+            STRCAT(contentType, MKV_H264_CONTENT_TYPE);
```
hassanctech, line 115: How did this happen, we added h265 much later....

hassanctech (approved):
LGTM

src/source/StreamInfoProvider.c
disa6302: Yeah. I was pretty surprised too!

src/source/Response.c
disa6302: As discused offline, added this back and commented it out.

hassanctech (dismissed):
lgtm

src/source/Response.h
```diff
@@ -108,14 +108,6 @@ STATUS freeCurlResponse(PCurlResponse*);
  */
 STATUS closeCurlHandles(PCurlResponse);
 
-/**
- * Debug prints statistics of the request/response
- *
- * @param - PCurlResponse - IN - Response object
- *
- */
-VOID dumpResponseCurlEasyInfo(PCurlResponse);
```
niyatim23, line 117: I think we should add this back and comment it out like the contents of the function in _Response.c_

disa6302: Already did it.

niyatim23 (approved):
Great work!

hassanctech (approved):
lgtm

PR #454, opened 2024-09-12
v1.5.3 Release

hassanctech:
### Build Changes
- Actually link against mbedtls for common lws build.
  - If we're using mbedtls instead of ssl we need to actually link against it when building kvsCommonLws, it is done properly for common curl, but was incorrect for Lws path.

### CI Changes
- Switched from macos-13-xlarge to macos-latest for M1 jobs since xlarge runners are no longer required to use an M1 runner.
- Switched away from using macos-latest for non-M1 jobs since macos-latest runners now use M1 processors.
- Installed pkgconfiglite for the Windows job in order to successfully build the SDK.
- Updated GCC version used for non-M1 Mac jobs to 14 to fix GCC-MacSDK compatibilty.

*Issue #, if available:* n/a

*What was changed?*
target_link_libraries now actually links mbedtls

*Why was it changed?*
previously we were not correctly linking mbedtls

*How was it changed?*
By adding the missing mbedtls library in the target_link_libraries step

Tested with and without change, making the change resolved the issue


CMakeLists.txt
```diff
@@ -289,6 +289,7 @@ if(BUILD_COMMON_LWS)
       set_target_properties(kvsCommonLws PROPERTIES VERSION ${KINESIS_VIDEO_PRODUCER_C_VERSION} SOVERSION ${KINESIS_VIDEO_PRODUCER_C_MAJOR_VERSION})
     endif()
     target_link_libraries(kvsCommonLws
+            ${PRODUCER_CRYPTO_LIBRARY}
             ${OPENSSL_CRYPTO_LIBRARY}
```
stefankiesz, line 293: Just want to leave a note that the existing `${OPENSSL_CRYPTO_LIBRARY}` and `${OPENSSL_SSL_LIBRARY}` should likely not be here since `${PRODUCER_CRYPTO_LIBRARY}` will cover both openssl and mbedtls cases. This can be changed and tested in a separate PR to /develop.

stefankiesz (dismissed):
Nice, lgtm pending CI passes.

unicornss (dismissed):

stefankiesz (dismissed):

unicornss (approved):

sirknightj (approved):

stefankiesz (approved):

PR #465, opened 2024-12-20
Add Support for Static Producer to Shared `kvssink` Builds

stefankiesz:
This PR corresponds to the [kvssink .so with static linkage to cproducer (#1179)](https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/pull/1216) Producer Cpp PR.

*What was changed?*
- Allow Producer Cpp builds to pass on the `CURL_POSITION_INDEPENDENT_CODE` CMake flag to this Producer C library to allow for building a static Curl library with a shared `kvssink` library.

*Why was it changed?*
- Curl does not set `CMAKE_POSITION_INDEPENDENT_CODE` nor set `-fPIC`, so this must be set when building it statically to be consumed by a shared library.

*How was it changed?*
- Added the passing of the `CURL_POSITION_INDEPENDENT_CODE` flag from Producer C CMakeLists to Curl CMakeLists to set the `CMAKE_POSITION_INDEPENDENT_CODE` flag.

Tested the `kvssink` build locally on Ubuntu and added a CI job for it.


sirknightj:
Handled in 497

PR #473, opened 2025-02-04
Add sample checks for Windows, Mac and Linux

sirknightj:
*Issue #, if available:*
* N/A

*What was changed?*
- Adding verification for the samples on Windows with the mkv dump env flag
- Fix an issue where the codec string isn't properly null-terminated in the samples. Using strcpy, or const char* should be fine in this case because the arguments (constant values) are known ahead of time and not susceptible to user or network input.
- Remove an extra section with the file path (twice) in the KvsVideoOnlyRealtimeStreamingSample.c sample

*Why was it changed?*
- To verify portability of the sample applications and SDK options.

*How was it changed?*
- The CI will build the samples and run them

- CI should be passing for these changes:

image
image

And for Mac/Linux:
image

### DIFF
--- a/samples/KvsAudioOnlyStreamingSample.c
+++ b/samples/KvsAudioOnlyStreamingSample.c
@@ -107,7 +107,7 @@ INT32 main(INT32 argc, CHAR* argv[])
 
     MEMSET(&data, 0x00, SIZEOF(SampleCustomData));
 
-    STRCPY(audioCodec, AUDIO_CODEC_NAME_AAC); // aac audio by default
+    SNPRINTF(audioCodec, SIZEOF(audioCodec), "%s", AUDIO_CODEC_NAME_AAC); // aac audio by default
 
 #ifdef IOT_CORE_ENABLE_CREDENTIALS
     PCHAR pIotCoreCredentialEndpoint, pIotCoreCert, pIotCorePrivateKey, pIotCoreRoleAlias, pIotCoreThingName;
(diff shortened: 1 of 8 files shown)


samples/KvsAudioOnlyStreamingSample.c
```diff
@@ -132,7 +132,7 @@ INT32 main(INT32 argc, CHAR* argv[])
 
     if (argc >= 5) {
         if (!STRCMP(argv[2], AUDIO_CODEC_NAME_ALAW)) {
-            STRNCPY(audioCodec, AUDIO_CODEC_NAME_ALAW, STRLEN(AUDIO_CODEC_NAME_ALAW));
+            STRCPY(audioCodec, AUDIO_CODEC_NAME_ALAW);
```
unicornss, line 135: SNPRINTF(audioCodec, sizeof(audioCodec), "%s", AUDIO_CODEC_NAME_AAC);

samples/KvsVideoOnlyRealtimeStreamingSample.c
```diff
@@ -95,13 +95,6 @@ INT32 main(INT32 argc, CHAR* argv[])
     sessionToken = GETENV(SESSION_TOKEN_ENV_VAR);
 #endif
 
-    MEMSET(frameFilePath, 0x00, MAX_PATH_LEN + 1);
```
sirknightj, line 98: Note to self: Removed this section since it's in there twice (line 128)

stefankiesz (approved):

unicornss (approved):

PR #476, opened 2025-03-05
Dual-stack endpoints support

sirknightj:
*Issue #, if available:*
* N/A

*What was changed?*
* Added dual-stack endpoint support (endpoint URL construction)
* Added IP version filter
* Speed up the ControlPlaneUrl test
* Add new CMake flag combinations test

*Why was it changed?*
* SDK support for [dual-stack endpoints](https://docs.aws.amazon.com/general/latest/gr/rande.html#dual-stack-endpoints)

*How was it changed?*
- To construct the correct URL, the SDK needs to know two things:
  1. Region (currently a parameter exists)
  2. Dual-stack or the legacy one (no parameter previously existed)
- Therefore, the `useLegacyEndpoint` parameter was added.
- Since these changes double the number of endpoint combinations, I refactored `constructControlPlaneUrl` to add more clarity on the different URL components and use a single string template instead of one string template for each branch.

> [!IMPORTANT]
> For backwards compatibility, the legacy endpoint (IPV4 only) will still be the default.

- For the IP filters, the SDK will pass the selected setting to curl.
- Documented the Endpoint provider and IP version provider chain in the Readme.
  - The chain was designed to allow builders maximum flexibility to enable and disable environment variable control (without needing to fork or patch the SDK).

- Added CMake combination tests to verify that the CMake fails if multiple of the mutually-exclusive flags were specified.
  - This test has been added to the CI and outputs a markdown summary
- Manually verified the `#error` directives by commenting out the CMake checks.
- Rewrote existing unit test to verify the URL construction.

> [!NOTE]
> This specific original test took about 10s to run on my setup. This pull request speeds up that test to run in under 1ms on my setup, in addition to doubling the number of combinations checked.


src/source/Response.c
```diff
@@ -180,6 +180,35 @@ STATUS initializeCurlSession(PRequestInfo pRequestInfo, PCallInfo pCallInfo, CUR
...
+#else
+    if (!IS_NULL_OR_EMPTY_STRING(GETENV(KVS_CURL_IPRESOLVE_V4_ENV_VAR))) {
+        curl_easy_setopt(pCurl, CURLOPT_IPRESOLVE, CURL_IPRESOLVE_V4);
+        DLOGD("Curl resolving IPV4 through environment variable");
+    } else if (!IS_NULL_OR_EMPTY_STRING(GETENV(KVS_CURL_IPRESOLVE_V6_ENV_VAR))) {
+        curl_easy_setopt(pCurl, CURLOPT_IPRESOLVE, CURL_IPRESOLVE_V6);
+        DLOGD("Curl resolving IPV6 through environment variable");
+    } else {
+        curl_easy_setopt(pCurl, CURLOPT_IPRESOLVE, CURL_IPRESOLVE_WHATEVER);
+        DLOGD("Curl resolving dual stack by default");
```
stefankiesz, line 201: **Note for others:** `CURLOPT_IPRESOLVE` is the default `CURLOPT_IPRESOLVE` value, so this maintains same behavior for the legacy endpoint as before when we were not setting the curl option at all.

src/source/Response.c
```diff
@@ -180,6 +180,35 @@ STATUS initializeCurlSession(PRequestInfo pRequestInfo, PCallInfo pCallInfo, CUR
...
+    }
+#endif
+
+#ifndef NDEBUG
+    // Only available in debug build
+    if (!IS_NULL_OR_EMPTY_STRING(GETENV(KVS_CURL_DEBUG_ENV_VAR))) {
+        curl_easy_setopt(pCurl, CURLOPT_VERBOSE, 1L);
```
stefankiesz, line 208: Can you please add a comment here describing the intentions one would have for using this mode

stefankiesz (commented):
Leaving some initial comments, still have to review the tests.

src/source/Response.c
sirknightj: Sure. This would add curl ultra-debug logs https://ec.haxx.se/usingcurl/verbose/index.html. The relevant part here is that it will also print out the IP address that it resolved and connected to.

stefankiesz (commented):
Nice, tests look good!

CMakeLists.txt
```diff
@@ -23,6 +23,15 @@ option(BUILD_OPENSSL_PLATFORM "If buildng OpenSSL what is the target platform" O
 option(LOCAL_OPENSSL_BUILD "Whether or not to use local OpenSSL build" OFF)
 option(CONSTRAINED_DEVICE "Change pthread stack size" OFF)
 
+# Endpoint selection - mutually exclusive
```
unicornss, line 26: Should one of these be ON, as they are mutually exclusive?

README.md
```diff
@@ -197,6 +202,23 @@ For video and audio: `createOfflineAudioVideoStreamInfoProviderWithCodecs()`
...
+1. The `controlPlaneUrl` parameter for `createAbstractDefaultCallbacksProvider`.
+2. Endpoint configuration CMake parameters: (`-DAWS_KVS_USE_LEGACY_ENDPOINT_ONLY=TRUE`, `-DAWS_KVS_USE_DUAL_STACK_ENDPOINT_ONLY=TRUE`)
+3. Environment variables: (`export AWS_USE_DUALSTACK_ENDPOINT=TRUE`)
+  - If `AWS_USE_DUALSTACK_ENDPOINT` is `TRUE` (case-insensitive), the dual-stack endpoint will be used.
+4. Otherwise, the legacy endpoint will be constructed.
+
+With 2, 3, and 4, the endpoint will be constructed based on the region provided to `createAbstractDefaultCallbacksProvider`.
+
+#### DNS filtering
+1. DNS resolution CMake parameters: (`-DAWS_KVS_IPV4_ONLY=TRUE`, `-DAWS_KVS_IPV6_ONLY=TRUE`, `-DAWS_KVS_DUAL_STACK_ONLY=TRUE`)
```
unicornss, line 218: suggest adding a note - "ONLY" in this context refers to the endpoint being enabled strictly or forced.

tst/ProducerCallbackProviderApiTest.cpp
```diff
@@ -42,53 +43,93 @@ TEST_F(ProducerCallbackProviderApiTest, AddProducerCallbackProvider_Returns_Vali
...
+    std::replace(region.begin(), region.end(), '-', '_');
+
+    return region + "_" + (param.isLegacyEndpoint ? "Legacy" : "New");
 }
 
+// Instantiate the test suite with multiple test cases
+INSTANTIATE_TEST_SUITE_P(ControlPlaneUrlTests, ControlPlaneUrlTest,
+                         ::testing::Values(
+                             // Legacy endpoints
+                             ControlPlaneUrlTestParam{"us-east-1", "https://kinesisvideo.us-east-1.amazonaws.com", TRUE},
```
unicornss, line 85: Would it be better to group them by region or category, e.g., isolated regions, CN (China), etc.?

src/source/CurlApiCallbacks.c
```diff
@@ -77,26 +78,21 @@ STATUS createCurlApiCallbacks(PCallbacksProvider pCallbacksProvider, PCHAR regio
...
-        } else if (0 == STRNCMP(pCurlApiCallbacks->region, AWS_CN_REGION_PREFIX, STRLEN(AWS_CN_REGION_PREFIX))) {
-            SNPRINTF(pCurlApiCallbacks->controlPlaneUrl, MAX_URI_CHAR_LEN, "%s%s.%s%s", CONTROL_PLANE_URI_PREFIX, KINESIS_VIDEO_SERVICE_NAME,
-                     pCurlApiCallbacks->region, CONTROL_PLANE_URI_POSTFIX_CN);
-        } else {
-            // Create a fully qualified URI
-            SNPRINTF(pCurlApiCallbacks->controlPlaneUrl, MAX_URI_CHAR_LEN, "%s%s.%s%s", CONTROL_PLANE_URI_PREFIX, KINESIS_VIDEO_SERVICE_NAME,
-                     pCurlApiCallbacks->region, CONTROL_PLANE_URI_POSTFIX);
-        }
+    // Construct Control Plane URL
+    if (IS_NULL_OR_EMPTY_STRING(controlPlaneUrl)) {
```
unicornss, line 82: can it be moved to 
 
 static EndpointType determineEndpointType() {
#if defined(AWS_KVS_USE_LEGACY_ENDPOINT_ONLY)
    return ENDPOINT_TYPE_LEGACY;
#elif defined(AWS_KVS_USE_DUAL_STACK_ENDPOINT_ONLY)
    return ENDPOINT_TYPE_DUAL_STACK;
#else
    // Check environment variable
    const char* envVar = GETENV(CONTROL_PLANE_USE_DUAL_STACK_ENDPOINT_ENV_VAR);
    BOOL isDualStack = (envVar != NULL) && (0 == STRNCMPI(envVar, "true", 4));
    return isDualStack ? ENDPOINT_TYPE_DUAL_STACK : ENDPOINT_TYPE_LEGACY;
#endif
}
e.g. 
typedef enum {
    ENDPOINT_TYPE_LEGACY,
    ENDPOINT_TYPE_DUAL_STACK
} EndpointType;

CMakeLists.txt
sirknightj: At most 1 can be on. 0 is acceptable

CMakeLists.txt
```diff
@@ -64,6 +73,58 @@ if(NOT CMAKE_BUILD_TYPE)
...
+endif()
+
+# Sum the enabled flags
+math(EXPR IPV_OPTIONS_ENABLED_COUNT
+        "${IPV4_ONLY_FLAG}+${IPV6_ONLY_FLAG}+${DUAL_STACK_ONLY_FLAG}")
+
+# Check for conflicts
+if(IPV_OPTIONS_ENABLED_COUNT GREATER 1)
+    message(FATAL_ERROR "Only one of AWS_KVS_IPV4_ONLY, AWS_KVS_IPV6_ONLY, or AWS_KVS_DUAL_STACK_ONLY can be enabled.")
+endif()
```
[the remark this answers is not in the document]
sirknightj: AWS_KVS_IPV4_ONLY, AWS_KVS_IPV6_ONLY and AWS_KVS_DUAL_STACK_ONLY all work on my mac with the AWS_KVS_USE_LEGACY_ENDPOINT_ONLY flag.

I did notice that it seems that on mac, the curl CURL_IPRESOLVE_V6 resolves an IPV4-IPV6 mapped address for the legacy endpoint, which is out of our control in this case. However, on Ubuntu, it failed as expected.

https://github.com/curl/curl/issues/13112

README.md
sirknightj: I didn't want to use "forced" in the readme.

I think I might rename `AWS_KVS_DUAL_STACK_ONLY` to `AWS_KVS_IPV4_OR_IPV6_ONLY`.

tst/ProducerCallbackProviderApiTest.cpp
sirknightj: I left the original unit test order:
1. Commercial regions
2. Govcloud/fips regions
3. China regions

src/source/CurlApiCallbacks.c
sirknightj: Ok

src/source/Response.c
```diff
@@ -180,6 +180,35 @@ STATUS initializeCurlSession(PRequestInfo pRequestInfo, PCallInfo pCallInfo, CUR
...
+#elif defined(AWS_KVS_IPV6_ONLY)
+    curl_easy_setopt(pCurl, CURLOPT_IPRESOLVE, CURL_IPRESOLVE_V6);
+    DLOGD("Curl resolving IPV6 through AWS_KVS_IPV6_ONLY");
+#elif defined(AWS_KVS_DUAL_STACK_ONLY)
+    curl_easy_setopt(pCurl, CURLOPT_IPRESOLVE, CURL_IPRESOLVE_WHATEVER);
+    DLOGD("Curl resolving dual stack through AWS_KVS_DUAL_STACK_ONLY");
+#else
+    if (!IS_NULL_OR_EMPTY_STRING(GETENV(KVS_CURL_IPRESOLVE_V4_ENV_VAR))) {
+        curl_easy_setopt(pCurl, CURLOPT_IPRESOLVE, CURL_IPRESOLVE_V4);
+        DLOGD("Curl resolving IPV4 through environment variable");
```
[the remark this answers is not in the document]
sirknightj: I was thinking about that one. In AWS SDK docs they have this: https://docs.aws.amazon.com/sdkref/latest/guide/feature-endpoints.html where they say it can be true or false. False is an acceptable value there, but setting it to false here would enable it (since it's defined).

stefankiesz (approved):

unicornss (approved):
new feature: control plane ipv6/ipv4 support (and dataplane)

PR #517, opened 2026-08-12
use computed back-off for emulated API calls

vabhasin:
*Issue #, if available:*
- N/A

*What was changed?*
- Added `callAfter` enforcement to all four emulated/caching API call paths in `CurlApiCallbacks.c`: `describeStreamCachingCurl`, `createStreamCachingCurl`, `getStreamingEndpointCachingCurl`, and `tagResourceCachingCurl`.

*Why was it changed?*
- When API call caching is active (`API_CALL_CACHE_TYPE_ALL`, the default for kvssink and all samples), the state machine's exponential backoff wait time is passed via `pServiceCallContext->callAfter`. The real curl handler threads already honor this by sleeping before making the HTTP call. However, when the call is emulated (cache hit), the result event was fired synchronously without checking `callAfter`, effectively discarding the computed backoff.
- During a network outage with a fresh endpoint cache, this caused the state machine to cycle rapidly through emulated Describe → emulated GetEndpoint → cached GetToken → failed PutMedia at DNS-failure speed (~50 cycles/s). While the control-plane calls themselves are no-ops, each cycle reaches PutMedia which spawns a real thread, initializes a TLS session, signs the request, and attempts DNS resolution before failing, creating unnecessary churn, lock contention against `putFrame`, and a potential concurrent reconnect at fleet scale.

*How was it changed?*
- Added the same `callAfter` check-and-sleep pattern used by the auth callbacks and real curl handlers to each emulated path, before the result event is fired. For `getStreamingEndpointCachingCurl`, the sleep is placed before the endpoint cache lock acquisition to avoid holding the mutex during the wait.

- Reproduced the tight retry loop locally (blocked DNS resolution during active PutMedia streaming with fresh endpoint cache). Before the fix: ~50 PutMedia attempts/s visible in DEBUG logs, retry counter climbing rapidly with computed waits of 10–19 s never applied. After the fix: retry attempts properly spaced at the computed backoff intervals.
- Existing unit tests pass (`cmake .. && make && ./tst/producerTest`).

### DIFF
--- a/src/source/CurlApiCallbacks.c
+++ b/src/source/CurlApiCallbacks.c
@@ -1113,6 +1113,7 @@ STATUS createStreamCachingCurl(UINT64 customData, PCHAR deviceName, PCHAR stream
     PCurlApiCallbacks pCurlApiCallbacks = (PCurlApiCallbacks) customData;
     PCallbacksProvider pCallbacksProvider = NULL;
     BOOL emulateApiCall = TRUE;
+    UINT64 currentTime;
 
     CHK(pCurlApiCallbacks != NULL && pCurlApiCallbacks->pCallbacksProvider != NULL && pServiceCallContext != NULL, STATUS_INVALID_ARG);
     pCallbacksProvider = pCurlApiCallbacks->pCallbacksProvider;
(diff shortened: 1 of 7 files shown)


sirknightj (approved):

src/source/CurlApiCallbacks.c
```diff
@@ -1130,6 +1131,14 @@ STATUS createStreamCachingCurl(UINT64 customData, PCHAR deviceName, PCHAR stream
         CHK(FALSE, retStatus);
     }
 
+    // Respect the callAfter to honor backoff wait time from the state machine retry strategy
+    if (pServiceCallContext->callAfter != 0) {
+        currentTime = pCallbacksProvider->clientCallbacks.getCurrentTimeFn(pCallbacksProvider->clientCallbacks.customData);
+        if (currentTime < pServiceCallContext->callAfter) {
+            THREAD_SLEEP(pServiceCallContext->callAfter - currentTime);
+        }
+    }
```
YPandas: Maybe we can collapse all the same code pieces to a helper function?

vabhasin: Good suggestion, done

YPandas (approved):

sirknightj (approved):

